PR #73, opened 2026-03-03
feat(memvault): add secure in-memory vault with docs

jithinkunjachan:
* Introduce MemVaultData for secure memory regions backed by mmap'd anonymous
memory, locked to RAM, and supporting read-only protection via mprotect.
* Includes destroy functionality for secure zeroing and unmapping.
* Add platform specific alloc and NoDump implementations for Darwin and Linux. 
* Provide unit and integration tests for vault creation, destruction, and read-only behavior.


internal/securemem/handler.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+func newVaultSession() *VaultSession {
+	return &VaultSession{
+		toPersist: make(map[string]struct{}),
+		vault:     NewVault(),
+	}
+}
+
+func (r *VaultSession) Put(name string, data []byte) error {
```
apatsap: Why do we need the (non persistent) Put here?

jithinkunjachan:
Closing as it is implemented

PR #103, opened 2026-04-09
fix: use kr http client in admin client

fabenan-f:
The kr HTTP client is now used for the admin client, consistent with the agent client. Additionally, the coverage report has been fixed to display the total coverage percentage.


apatsap (approved):

jithinkunjachan (approved):
👍

PR #104, opened 2026-04-10
feat: add registry store for tracking agent registrations

jithinkunjachan:
Introduce a PostgreSQL-backed registry store that records agent instances and their heartbeat status during registration.
The agent API now requires an X-Agent-Id header and upserts registry entries on each registration request.


pkg/store/sql/registry_test.go
```diff
@@ -0,0 +1,198 @@
...
+			Status:     spec.RegistryStatusHealthy,
+		}
+
+		// when
+		upsertResult, err := subj.Upsert(ctx, store.UpsertRegistryQuery{
+			Registry: registry,
+		})
+
+		// then
+
```
apatsap: NP.: the new lines after //given // when // then are not uniformly accross the test. sometimes they have an empty line sometimes they haven't. Pls make it uniform (I'd vote for no blank line)

internal/spec/registry.go
```diff
@@ -0,0 +1,22 @@
+package spec
```
apatsap: Why is this in the spec? This thing doesn't describe/configures anything?

internal/spec/registry.go
```diff
@@ -0,0 +1,22 @@
...
+type RegistryStatus string
+
+// RegistryStatusHealthy indicates that the registry is active and healthy.
+const (
+	RegistryStatusHealthy   RegistryStatus = "healthy"
+	RegistryStatusUnhealthy RegistryStatus = "unhealthy"
+)
+
+// Registry entries are used to track the status of agents and their heartbeats.
+type Registry struct {
```
apatsap: Its not a `Registry` but more an `AgentRegistration` or `AgentRecord`

pkg/store/registry.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/openkcm/krypton/internal/spec"
+)
+
+var ErrRegistryNotFound = errors.New("registry not found")
+
+type Registry interface {
```
apatsap: This is the `AgentRegistry`. `Registry` is far too broad because, this thing will only have the AgentRegistration related stuff

apatsap (changes requested):
Nice. Just some naming remarks

apatsap (approved):
Turned out nicely!

fabenan-f (approved):
looks good!

PR #107, opened 2026-04-10
Feat/add simple worker

jithinkunjachan:
Add worker scheduler with a Stop method for explicit shutdown,
pass context to task functions, ensure Start is only effective once,
and validate that the interval is positive. 

Shall we add a Name for the Scheduler.

### DIFF
--- a/internal/worker/worker.go
+++ b/internal/worker/worker.go
@@ -1,0 +1,78 @@
+// Package worker provides a simple periodic task scheduler.
+package worker
+
+import (
+	"context"
+	"errors"
+	"log/slog"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// TaskFn is a function executed by a [Scheduler] on each tick.
+type TaskFn func(context.Context) error
+
+// Scheduler runs a [TaskFn] at a fixed interval until stopped via context cancellation or [Scheduler.Stop].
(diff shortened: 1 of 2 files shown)


apatsap (dismissed):

internal/worker/worker.go
```diff
@@ -0,0 +1,78 @@
...
+			return
+		case <-w.done:
+			return
+		}
+	}
+}
+
+// Stop signals the scheduler to stop. It is safe to call multiple times.
+func (w *Scheduler) Stop() {
+	w.stopOnce.Do(func() {
```
apatsap, line 75: the stop isn't reflected in `isStarted`.

So you can either reflect it in `isStarted` or call this method `Kill`, to imply that it can't be re-started

apatsap (changes requested):
found one small thing

apatsap (approved):

internal/worker/worker_test.go
```diff
@@ -0,0 +1,250 @@
+package worker_test
+
+import (
+	"context"
+	"sync/atomic"
+	"testing"
+	"testing/synctest"
```
fabenan-f, line 7: nice!

fabenan-f (approved):

PR #110, opened 2026-04-14
feat: CLI output select method

fabenan-f:
This adds a select method to the CLI output builder. The provided function receives `output.Rows` and returns an index.


apatsap (approved):

jithinkunjachan (approved):
Good

PR #111, opened 2026-04-14
feat: add agent deregister endpoint with client

jithinkunjachan:
This PR adds deregister endpoint and client

### DIFF
--- a/pkg/api/agents/client.go
+++ b/pkg/api/agents/client.go
@@ -51,22 +51,9 @@ func NewClient(baseURL, agentName, agentID string) (*Client, error) {
 
 // Register sends an agent registration request to the server.
 func (c *Client) Register(ctx context.Context, req RegisterRequest) (RegisterResponse, error) {
-	body, err := json.Marshal(req)
+	httpResp, err := c.doCall(ctx, PathRegister, req)
 	if err != nil {
-		return RegisterResponse{}, fmt.Errorf("%w: %w", api.ErrFailedToEncodeRequest, err)
-	}
-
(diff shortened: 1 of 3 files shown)


pkg/api/agents/client.go
```diff
@@ -120,3 +120,36 @@ func (c *Client) SendHeartbeat(ctx context.Context, req SendHeartbeatRequest) (S
...
+	case http.StatusOK:
+		var resp DeregisterResponse
+		err = json.NewDecoder(httpResp.Body).Decode(&resp)
+		if err != nil {
+			return DeregisterResponse{}, fmt.Errorf("%w: %w", api.ErrFailedToDecodeResponse, err)
+		}
+		return resp, nil
+	default:
+		return DeregisterResponse{}, fmt.Errorf("%w: expected %d Created, got %d", api.ErrUnexpectedStatusCode, http.StatusOK, httpResp.StatusCode)
+	}
```
apatsap: This pretty much 90% the same as the other methods. I guess its worth to DRY it a bit by adding a `doCall(....)` method (or something along those lines) and use it in all three methods

apatsap (changes requested):
Its a solid client, but I think we can DRY it a bit

pkg/api/agents/client.go
jithinkunjachan: ~~yes I agree same is true for the handler. this refactoring will be done in the subsequent PR.~~
Good point refactored in the same PR

apatsap (approved):

fabenan-f (approved):

PR #114, opened 2026-04-15
add key hierarchy and topology cross validation

NamanBalaji:
- Add cross validation between key hierarchy, segments, and topology that catches configuration errors at load time 
  - Enforce root segment rules (must start at hierarchy root, cannot end on tek, root key cannot have parent provider) and agent segment rules (must start on tek, tek must have parent provider, cannot contain root key)
  - Validate topology level : no duplicate segment names, all ParentKeyProvider references resolve to known agents, and referenced parent agents actually manage the parent key kind

### DIFF
--- a/internal/spec/hierarchy_validation.go
+++ b/internal/spec/hierarchy_validation.go
@@ -27,15 +27,14 @@ var (
 	ErrParentKeyProviderKeyNotInParentSegment = errors.New("parent key provider's segment does not contain the parent key kind")
 )
 
-// ValidateSegmentAgainstHierarchy checks that a segment's endpoints exist in the hierarchy are in the correct order,
-// and that bindings cover exactly the segment range.
+// ValidateSegmentAgainstHierarchy checks that a segment's endpoints exist in the hierarchy are in the correct order, and that bindings cover exactly the segment range.
 func ValidateSegmentAgainstHierarchy(h KeyHierarchy, seg HierarchySegment, bindings map[string]KeyBinding) error {
-	startIdx, ok := h.IndexOf(KeyKind(seg.StartKind))
-	if !ok {
+	startIdx := h.IndexOf(KeyKind(seg.StartKind))
+	if startIdx < 0 {
 		return fmt.Errorf("%w: %q", ErrSegmentStartKindNotInHierarchy, seg.StartKind)
 	}
-	endIdx, ok := h.IndexOf(KeyKind(seg.EndKind))
-	if !ok {
+	endIdx := h.IndexOf(KeyKind(seg.EndKind))
+	if endIdx < 0 {
 		return fmt.Errorf("%w: %q", ErrSegmentEndKindNotInHierarchy, seg.EndKind)
 	}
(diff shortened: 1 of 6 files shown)


internal/spec/hierarchy_validation.go
```diff
@@ -0,0 +1,193 @@
...
+	if !exists || startBinding.ParentKeyProvider == nil {
+		return ErrAgentSegmentStartMissingParentKeyProvider
+	}
+
+	return nil
+}
+
+// ValidateTopologyAgainstHierarchy validates the entire topology against the hierarchy checking agent segment rules,
+// name uniqueness, and parent key provider resolution.
+func ValidateTopologyAgainstHierarchy(h KeyHierarchy, topology Topology, rootSeg HierarchySegment, rootBindings map[string]KeyBinding) error {
```
apatsap: NP:
we have `h` (fine with me since the type is explicit and can't be broken down to smaller units.
and we have `topology` which is also has an very explicit type. So this should be either `t` or `top`

internal/spec/hierarchy_validation.go
```diff
@@ -0,0 +1,193 @@
...
+	segmentMap := make(map[string]TopologySegment, len(topology.Segments))
```
apatsap: this is technically a set where we only care about the keys. So I'd make it 

`uniqSegs :=  make(map[string]struct{}, len(topology.Segments))`

internal/spec/hierarchy_validation.go
```diff
@@ -0,0 +1,193 @@
...
+			continue
+		}
+		providerName := binding.ParentKeyProvider.AgentName
+		providerSeg, exists := providers[providerName]
+		if !exists {
+			return fmt.Errorf("%w: %q (referenced by key kind %q)", ErrParentKeyProviderNotResolvable, providerName, kind)
+		}
+
+		kindIdx, ok := h.IndexOf(KeyKind(kind))
+		if !ok || kindIdx == 0 {
```
apatsap: Ususally, `Index` (`slices`,`string`) methods in go only return an int. and a `-1` when the element wasn't found. So I would do the same here. It also makes the code a bit more crisp

internal/spec/keyhierarchy.go
```diff
@@ -90,3 +94,33 @@ func (h KeyHierarchy) FindKeySpec(kind KeyKind) (KeySpec, bool) {
 	}
 	return KeySpec{}, false
 }
+
+// IndexOf returns the position of the given kind in the hierarchy's KeySpecs slice.
+// Returns the index and true if found, or -1 and false if not found.
+func (h KeyHierarchy) IndexOf(kind KeyKind) (int, bool) {
```
apatsap: I'd only make it return an `int` (see above comment)

apatsap (changes requested):
just some small nps

apatsap (approved):

fabenan-f (approved):

jithinkunjachan (approved):
LGTM

PR #116, opened 2026-04-15
feat(cli): store tenant selection in state file

fabenan-f:
Add local storage for selected tenant in `~/.krypton/state.lock`, allowing the CLI to persist and reuse tenant selection across commands.


cli/config/config.go
```diff
@@ -0,0 +1,110 @@
...
+var (
+	// ErrConfigNotFound is returned when the configuration file does not exist.
+	ErrConfigNotFound = errors.New("config not found")
+	// ErrConfigNil is returned when attempting to save a nil configuration.
+	ErrConfigNil = errors.New("config cannot be nil")
+)
+
+const (
+	Directory      = ".krypton"
+	ConfigFileName = "config.json"
```
apatsap: I think config is a bit misleading here. A config for me is an instruction set that describes how a software should run/is configured. 

So I'd go with something like `krypton.lock` or something like `state.lock`, because we don't want the user to edit this file

cli/config/config.go
```diff
@@ -0,0 +1,110 @@
...
+
+// TenantSelection stores the selected tenant information.
+type TenantSelection struct {
+	// ID is the unique identifier of the tenant.
+	ID string `json:"id"`
+	// Name is the display name of the tenant.
+	Name string `json:"name,omitempty"`
+}
+
+// Store persists configuration to ~/.krypton/config.json.
```
apatsap: NP putting this is `~/.krypton/config.json.` into the comment is mild technical debt, because whenever we change path we need to change all comments.

cli/config/config.go
```diff
@@ -0,0 +1,110 @@
...
+type Store struct {
+	dirPath    string
+	configPath string
+}
+
+// NewStore creates a new store that persists configuration to ~/.krypton/config.json.
```
apatsap: ~/.krypton/config.json.

see above

cli/config/config.go
```diff
@@ -0,0 +1,110 @@
...
+	ErrConfigNil = errors.New("config cannot be nil")
+)
+
+const (
+	Directory      = ".krypton"
+	ConfigFileName = "config.json"
+)
+
+// Config represents local CLI configuration.
+type Config struct {
```
apatsap: I think this is a state or selection not really a config

apatsap (changes requested):
mainly just some wording

cli/config/config.go
fabenan-f: Good point, config is really misleading here. I went with `.krypton/state.lock`

apatsap (approved):

jithinkunjachan (approved):
LGTM

PR #118, opened 2026-04-20
feat: replace cmd/server with cmd/root and cmd/agent, consolidate con…

jithinkunjachan:
- Replaced cmd/server with separate cmd/root (root server) and cmd/agent (agent binary) entry points
- The agent binary implements the full lifecycle: register, heartbeat, and graceful deregister on shutdown
- Made NewServerMux composable -- both admin and agents packages now accept/return *http.ServeMux for chaining
- Consolidated hierarchy + topology params into a single config.RootConfig in the agents handler
- Root server includes a worker to monitor agent health (unhealthy after 30s, deregistered after 90s)
- Updated Makefile with root and agent targets
- Updated all tests to match new signatures

# Small demo for agent registration

https://github.com/user-attachments/assets/3c40e2e1-ee2c-416b-898f-5bc04dc51a50


cmd/agent/main.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/google/uuid"
+
+	"github.com/openkcm/krypton/internal/worker"
+	"github.com/openkcm/krypton/pkg/api/agents"
+)
+
+// This is a simple agent that registers itself with the root server,
+// sends periodic heartbeats, and deregisters on shutdown.
+func main() {
+	rootSrvPort := os.Getenv("ROOT_SERVER_PORT")
```
fabenan-f: We could also "load" `config.AgentBootstrapConfig` here like the sample config for root

pkg/api/agents/handler.go
```diff
@@ -32,9 +33,8 @@ const (
 )
 
 type agent struct {
-	store     store.Agent
-	hierarchy spec.KeyHierarchy
-	topology  spec.Topology
+	store store.Agent
+	root  config.RootConfig
```
fabenan-f: np: the field name `root` in combination with `Hierarchy` sounds like it is only about the hierarchy concerning root. I think `config` fits better

fabenan-f (approved):
just two things which may be considered or dismissed

fabenan-f (approved):

cmd/agent/main.go
```diff
@@ -0,0 +1,87 @@
...
+		if port == "" {
+			port = ":8080"
+		}
+		return &config.AgentBootstrapConfig{
+			Name: "agent-k1",
+			Role: spec.DefaultRole,
+			KryptonRoot: config.KryptonRoot{
+				Address: config.Address{
+					Type: config.AddressTypeHTTP,
+					URL:  "http://localhost" + port,
```
apatsap: `ROOT_SERVER_PORT` implies to me `1234` so this will fail here due to the missing `:`

cmd/agent/main.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+func loadConfig() *config.AgentBootstrapConfig {
+	cfg, err := config.LoadAgentBootstrapConfig(os.Getenv("AGENT_BOOTSTRAP_CONFIG_PATH"))
+	if err != nil {
+		port := os.Getenv("ROOT_SERVER_PORT")
+		if port == "" {
+			port = ":8080"
+		}
```
apatsap, line 74:
```
else {
_,err := strconv.Atoi(port)
```
Will fail faster

cmd/agent/main.go
```diff
@@ -0,0 +1,87 @@
...
+		log.Fatalf("%s: %v", msg, err)
+	}
+}
+
+func loadConfig() *config.AgentBootstrapConfig {
+	cfg, err := config.LoadAgentBootstrapConfig(os.Getenv("AGENT_BOOTSTRAP_CONFIG_PATH"))
+	if err != nil {
+		port := os.Getenv("ROOT_SERVER_PORT")
+		if port == "" {
+			port = ":8080"
```
apatsap: port is string rep as int. so `:` should go

cmd/root/main.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/openkcm/krypton/pkg/store"
+	storesql "github.com/openkcm/krypton/pkg/store/sql"
+)
+
+// Simple krypton server for manual testing and development.
+// Not intended for production use (yet).
+func main() {
+	srvPort := os.Getenv("SERVER_PORT")
+	if srvPort == "" {
+		srvPort = ":8080"
```
apatsap: Lets have the convention that port is always an int. No `:`

Makefile
```diff
@@ -63,9 +63,15 @@ postgres:
 postgres-stop:
 	docker rm -f $(POSTGRES_CONTAINER) 2>/dev/null || true
 
-.PHONY: server
-server:
-	DATABASE_URL="$(DATABASE_URL)" go run ./cmd/server
+ROOT_SERVER_PORT := :8080
```
apatsap: `:` should go

apatsap (changes requested):
Very nice. Demo video is super nice.

Just some remarks around port. My proposal is to always expect a string representation of int

apatsap (approved):

PR #122, opened 2026-04-21
feat: add integration test for agent registration lifecycle

jithinkunjachan:
This covers the full register/deregister/delete flow with dynamic port allocation, explicit subprocess environments, and coverage support.


apatsap (approved):
Nice and compact. We can add the unhappy cases later.

fabenan-f (approved):

PR #124, opened 2026-04-23
feat: add gRPC agent service with proto definition and tests

jithinkunjachan:
Implement AgentService with Register, SendHeartbeat, and Deregister RPCs, including proto/spec conversion layer and integration tests.


api-specs/v1/proto/agents/agents.proto
```diff
@@ -0,0 +1,82 @@
...
+
+message SendHeartbeatResponse {
+}
+
+message DeregisterAgentRequest {
+  string agent_name = 1;
+  string instance_id = 2;
+}
+
+message DeregisterAgentResponse {
```
jithinkunjachan, line 36: Do we need a this `bool success = 1`

https://github.com/openkcm/api-sdk/blob/a5894daec262fb0538d7d81cd7c09065e2b6ba6e/proto/kms/api/cmk/registry/system/v1/system.proto#L116-L118
```proto

message RemoveSystemLabelsResponse {
  bool success = 1;
}
```

pkg/api/agents/v1/proto/agents.proto
```diff
@@ -0,0 +1,82 @@
...
+message KeyBinding {
+  VaultSpec vault = 1;
+  ParentKeyProviderRef parent_key_provider = 2; // optional in proto3 by nature
+  map<string, string> labels = 3;
+}
+
+message VaultSpec {
+  string name = 1;
+  string type = 2;
+  google.protobuf.Struct params = 3;
```
jithinkunjachan: used this `google.protobuf.Struct` as this is a `map[string]any`

this will not support slices and int

pkg/api/agents/v1/proto/agents.proto
```diff
@@ -0,0 +1,82 @@
...
+
+message KeyHierarchy {
+  string name = 1;
+  repeated KeySpec key_specs = 2;
+}
+
+message KeySpec {
+  string kind = 1;
+  string role = 2;
+  string algorithm = 3;
```
jithinkunjachan: Did not use enums for `algorithm`, `role` do we need to use as this is already validated configuration?

api-specs/v1/proto/agents/agents.proto
apatsap: Well it would help us to distinguish between a transport error and an upstream error. In both cases we need to re-try so from this perspective it doesn't matter. But I'd still keep it for former distinction, but without having an actual hard argument why.

pkg/api/agents/v1/proto/agents.proto
apatsap: Both needs to be enum because, they can't be arbitrary

pkg/api/agents/v1/proto/agents.proto
```diff
@@ -0,0 +1,82 @@
...
+}
+
+message KeyHierarchy {
+  string name = 1;
+  repeated KeySpec key_specs = 2;
+}
+
+message KeySpec {
+  string kind = 1;
+  string role = 2;
```
apatsap: pls enum see below

pkg/api/agents/v1/proto/agents.proto
apatsap: pls enum see above

pkg/api/agents/v1/proto/agents.proto
```diff
@@ -0,0 +1,82 @@
...
+  HierarchySegment segment = 3;
+  map<string, string> labels = 4;
+  string role = 5;
+  KeyHierarchy hierarchy = 6;
+  int32 keep_alive = 7;
+}
+
+message KeyBinding {
+  VaultSpec vault = 1;
+  ParentKeyProviderRef parent_key_provider = 2; // optional in proto3 by nature
```
apatsap: I don't know what this comment tries to convey. pls rm

pkg/api/agents/v1/convert.go
```diff
@@ -0,0 +1,235 @@
...
+		KeyBindings: keyBindings,
+		Segment:     hierarchySegmentToProto(cfg.Segment),
+		Labels:      map[string]string(cfg.Labels),
+		Role:        string(cfg.Role),
+		Hierarchy:   keyHierarchyToProto(cfg.Hierarchy),
+		KeepAlive:   int32(cfg.KeepAlive),
+	}, nil
+}
+
+// --- Proto -> Spec helpers ---
```
apatsap: pls rm

pkg/api/agents/v1/convert.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	return spec.KeySpec{
+		Kind:      spec.KeyKind(pb.GetKind()),
+		Role:      spec.KeyRole(pb.GetRole()),
+		Algorithm: spec.KeyAlgorithm(pb.GetAlgorithm()),
+	}
+}
+
+// --- Spec -> Proto helpers ---
```
apatsap: pls rm

pkg/api/agents/v1/service.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func NewAgentService(store store.Agent, config config.RootConfig) *AgentService {
+	return &AgentService{
+		store:  store,
+		config: config,
+	}
+}
+
+// Register implements [proto.AgentServiceServer].
```
apatsap: Line 18 tells us that already. pls rm

pkg/api/agents/v1/service.go
```diff
@@ -0,0 +1,150 @@
...
+		slog.Error("failed to register agent", "agentName", agentName, "instanceID", instanceID, "error", err)
+		return nil, status.Error(codes.Internal, "failed to register agent")
+	}
+
+	return &proto.RegisterAgentResponse{
+		Config: pCfg,
+	}, nil
+}
+
+// SendHeartbeat implements [proto.AgentServiceServer].
```
apatsap: pls rm. see above

pkg/api/agents/v1/service.go
```diff
@@ -0,0 +1,150 @@
...
+	})
+	if err != nil {
+		slog.Error("failed to update agent heartbeat", "agentName", agentName, "instanceID", instanceID, "error", err)
+		return nil, status.Error(codes.Internal, "failed to update heartbeat")
+	}
+
+	return &proto.SendHeartbeatResponse{}, nil
+}
+
+// Deregister implements [proto.AgentServiceServer].
```
apatsap: pls rm. see above

pkg/api/agents/v1/proto/agents.proto
apatsap: I think this this limitation is not good. also we are potentially loosing the type of slices (bc I think `[]any` is supported)

apatsap: Let's zoom out.

The problem arises from having 
```go
type VaultSpec struct{
					Name string
					Type string
					Params map[string]any <- THIS is the culprit
}
```
Having `Params: map[string]any` is super flexibel but also quite generic, which makes it harder to configure for any user (bad).

Furthermore, each agent/root needs to be able to connect to a vault using something from its binary (so it isn't that arbitrary as the params type suggests). Given all the problems (proto limits, non-explicit config,a lot of boiler plate code for safe type-casting), we can make specific.

Idea:

```go
type VaultSpec struct{
					Name string
					Type string
					Config VaultConfig,
}

type VaultConfig interface {
     IsValidVaultConfig()error // I know not elegant since Validate()error would be better, but we need to avoid interface collision, so we need to make sure nothing implements this
} 

// implements VaultConfig
type AWSConfig struct {
    Region    string
    AccessKey string
    SecretKey string
}
// implements VaultConfig
type OpenBaoVaultConfig struct {
    Address   string
    Token     string
    MountPath string
}
```

In proto we would then get a `oneof` field that mirrors the Go interface:
```proto
message VaultSpec {
  string name = 1;
  
  oneof config {
    AWSConfig      aws      = 2;
    OpenBaoConfig  open_bao = 3;
  }
}

message AWSConfig {
  string region     = 1;
  string access_key = 2;
  string secret_key = 3;
}

message OpenBaoConfig {
  string address    = 1;
  string token      = 2;
  string mount_path = 3;
}
```

Adavantage safely typed, validation and if we ever allow dynamic vault configs much easier to grasp interface (no generic map, where the user needs to check a doc what to put where)

apatsap (changes requested):
Thanks for driving out the problem, which is due to a lazy/too generic go struct design. I've added an idea, which you are free to improve

api-specs/v1/proto/agents/agents.proto
jithinkunjachan: This will be tackled in the next PRs

pkg/api/agents/v1/proto/agents.proto
jithinkunjachan: This will be tackled in the next [PR](https://github.com/openkcm/krypton/pull/127) and we are going for yaml bytes

apatsap (approved):

pkg/api/agents/v1/service.go
```diff
@@ -0,0 +1,154 @@
+package v1
```
fabenan-f: shall we go with the version or for every version use the persona like `admin` or `agents` as the package name?

fabenan-f (approved):
just one question which might not affect the approval depending on the answer

pkg/api/agents/v1/service.go
jithinkunjachan: Good question or shall we go for `package adminv1` and  `package agentsv1` got some inspiration from  [api-sdk ](https://github.com/openkcm/api-sdk/blob/main/proto/kms/api/cmk/registry/auth/v1/auth.pb.go)

fabenan-f (approved):

apatsap (approved):
Nice

jithinkunjachan:
thank you everyone for having a good  discussion 👍

PR #125, opened 2026-04-24
feat: add Krypton identity model and policy-based authorization engine

fabenan-f:
Introduces two new packages:
- `identity`: definition of the identity model plus selector mechanism with glob-matching used in authz
- `authz`: policy-based authorization engine with topology-derived builtin policies


internal/spec/identity.go
```diff
@@ -0,0 +1,121 @@
...
+	// ErrInvalidIdentityRole is returned when the role is not 'root' or 'agent'.
+	ErrInvalidIdentityRole = errors.New("invalid identity role: must be 'root' or 'agent'")
+	// ErrEmptyAgentName is returned when agent name is empty.
+	ErrEmptyAgentName = errors.New("agent name cannot be empty")
+	// ErrNoKryptonID is returned when certificate has no krypton identity.
+	ErrNoKryptonID = errors.New("no krypton identity in certificate")
+)
+
+// KryptonID represents a krypton identity.
+type KryptonID struct {
```
apatsap: I guess we can just call it `Identity`. Reasons" because everything in this repo is krypton and `Identity` implies something more complex than just `ID`

internal/spec/identity.go
```diff
@@ -0,0 +1,121 @@
...
+	TrustDomain string    `json:"trust_domain"`
+	Role        AgentRole `json:"role"`
```
apatsap: I guess we can just call it Role instead of AgentRole because then we can also re-use it for authorization outside the agent domain

internal/spec/identity.go
```diff
@@ -0,0 +1,121 @@
...
+	Name        string    `json:"name"`
+}
+
+// URI returns the URI representation.
+// Format: kryptonid://<trust-domain>/<role>/<name>
+func (k KryptonID) URI() string {
+	if k.Role == RootRole {
+		return fmt.Sprintf("%s://%s/%s", KryptonIDScheme, k.TrustDomain, RootRole)
+	}
+	return fmt.Sprintf("%s://%s/%s/%s", KryptonIDScheme, k.TrustDomain, DefaultRole, k.Name)
```
apatsap: why does the first one not have the format from the comment?

internal/spec/identity.go
jithinkunjachan: If above is true then could you please also change the `DefaultRole` to `AgentRole`

internal/spec/identity.go
fabenan-f: The comment is not clear enough here. Since there can only be one root, there is no need for a name differentiator, so the path ends after `kryptonid://<trust-domain>/root`

internal/authz/glob.go
apatsap, line 1: why do we call this glob?

internal/identity/identity.go
```diff
@@ -0,0 +1,136 @@
...
+type Identity struct {
+	TrustDomain string
+	Segments    []Segment
+}
+
+// Segment represents a kind/name pair in the identity path.
+type Segment struct {
+	Kind string
+	Name string
+}
```
apatsap, line 57: Why not creating an own pkg. 

`/pkg/identity`

that holds:
* `identity.Identity`
* `identity.Segment`

It would also read better:

`identity.Parse`  instead of `authn.Parse`

internal/authn/identity.go
```diff
@@ -0,0 +1,136 @@
...
+			return Identity{}, err
+		}
+		segments = append(segments, Segment{Kind: parts[i], Name: parts[i+1]})
+	}
+
+	return Identity{TrustDomain: td, Segments: segments}, nil
+}
+
+// NodeIdentity creates a node identity for the given trust domain and name.
+func NodeIdentity(trustDomain, name string) (Identity, error) {
```
apatsap: What does the `NodeIdentity` actually mean or does?

internal/authn/identity.go
```diff
@@ -0,0 +1,136 @@
...
+		TrustDomain: trustDomain,
+		Segments: []Segment{
+			{Kind: KindNode, Name: name},
+		},
+	}, nil
+}
+
+// validateSegment checks that kind and name are non-empty and do not contain
+// glob characters ('*'), slashes, or whitespace.
+func validateSegment(kind, name string) error {
```
apatsap: can't we make this a method.

internal/authz/builtin.go
```diff
@@ -0,0 +1,116 @@
...
+				},
+			},
+		})
+	}
+
+	return policies, nil
+}
+
+// toPatternSegments converts concrete authn.Segments to PatternSegments (exact match).
+func toPatternSegments(segs []authn.Segment) []PatternSegment {
```
apatsap: I don't really understand the split here. why we have 2 patterns.

internal/authz/glob.go
```diff
@@ -0,0 +1,182 @@
...
+	// ErrEmptyPatternTrustDomain is returned when pattern trust domain is empty.
+	ErrEmptyPatternTrustDomain = errors.New("pattern trust domain cannot be empty")
+
+	// ErrInvalidPatternPath is returned when the pattern path is malformed.
+	ErrInvalidPatternPath = errors.New("invalid pattern path: must have kind/name pairs, optionally ending with **")
+)
+
+// IdentityPattern is a structured pattern for matching against authn.Identity values.
+// TrustDomain is always literal (no wildcards).
+type IdentityPattern struct {
```
apatsap: I think this is more a selector as a `Selector`. Because we currently do `IdentityPattern.Matches` and i think an `identity.Selctor.Matches()` is a little bit more crisp

apatsap (commented):
I like it. I think we can streamline the structure a bit

we have identity related stuff both in autn and authz with some having an overlap (see pattern conversion).

I think we should split it as follows

`/internal/identity` 
   * has the identity
   * all the primitives
`/internal/authn` 
 * will have all coneptually the `Authenticator` interface  x509Authenticator.Validate(identity) 
`/internal/authz` 
* will have the policy stuff you've build

internal/authz/glob.go
fabenan-f: lack of naming skills in this case

internal/authn/identity.go
fabenan-f: If a node starts and gets to know its domain and name, it can conveniently receive its identity with this func. But I removed it since this is not really relevant for this draft

internal/authz/builtin.go
fabenan-f: The split between `Segment` and `SegmentSelector` (here still `PatternSegment`)? `SegmentSelector` allows wildcards while `Segment` only concrete values

internal/authz/glob.go
fabenan-f: I was hoping for such a naming suggestion, I like it way better

internal/authz/builtin.go
apatsap: Gotcha. I think then the `Selector` makes even more sense.

PR #131, opened 2026-05-05
added reconciler pacakge with manager creation logic

NamanBalaji:


internal/reconciler/manager.go
```diff
@@ -0,0 +1,222 @@
...
+// TargetClientFactory builds Orbital initiators from reconciler targets.
+type TargetClientFactory interface {
+	NewInitiator(ctx context.Context, target config.ReconcilerTarget) (orbital.Initiator, error)
+}
+
+type TargetClientFactoryFunc func(context.Context, config.ReconcilerTarget) (orbital.Initiator, error)
+
+func (f TargetClientFactoryFunc) NewInitiator(ctx context.Context, target config.ReconcilerTarget) (orbital.Initiator, error) {
+	return f(ctx, target)
+}
```
apatsap: We have `TargetClientFactory` and `TargetClientFactoryFunc`.

Can't we make it more simplistic and just use:

`type TargetProvider func(context.Context, config.ReconcilerTarget) (orbital.Initiator, error)`

Because technically `TargetClientFactoryFunc` is exactly that just wrapped in another interface.

apatsap (changes requested):
just one simplification

internal/reconciler/manager.go
NamanBalaji: Yes makes sense we just have targets of one type so it's overcomplicating

apatsap (approved):

jithinkunjachan (approved):

internal/reconciler/manager.go
```diff
@@ -0,0 +1,213 @@
...
+	if targetProvider == nil {
+		return nil, ErrTargetFactoryRequired
+	}
+
+	for _, targetConfig := range targetConfigs {
+		client, err := targetProvider(ctx, targetConfig)
+		if err != nil {
+			return nil, errors.Join(fmt.Errorf("create target %s client: %w", targetConfig.Name, err), closeTargets(ctx, targets))
+		}
+		if client == nil {
```
jithinkunjachan, line 192: Just a small thing will we return a nil without an error?

PR #132, opened 2026-05-08
feat: add base key api and announce key req

NamanBalaji:


api-specs/v1/proto/keys/keys.proto
```diff
@@ -0,0 +1,46 @@
+syntax = "proto3";
+
+
+package krypton.v1.keys;
```
fabenan-f: I was wondering if we don't want to keep the packages organized by consumer type meaning moving the `KeyService` to the `krypton.v1.admin` package. This would also mean we need to prefix the existing `Service` with `Tenant`. This would group admin-related operations together and make the API more intuitive for consumers

pkg/api/v1/proto/admin/admin_grpc.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
-// - protoc-gen-go-grpc v1.6.1
+// - protoc-gen-go-grpc v1.5.1
```
fabenan-f: We should use the latest version if possible

pkg/model/key.go
```diff
@@ -0,0 +1,44 @@
+package model
+
+import (
+	"github.com/google/uuid"
+
+	"github.com/openkcm/krypton/internal/clock"
+)
+
+type KeyState string
```
fabenan-f, line 11: I think the intention was to have a processing state like this state and a key (lifecycle) state which is already defined by the `keylifecycle.State`, so I'd rename this to be more precise

pkg/store/sql/key.go
```diff
@@ -0,0 +1,93 @@
...
+		&key.UpdatedAt,
+	)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, store.ErrKeyNotFound
+		}
+		return nil, err
+	}
+
+	key.Kind = kind
```
jithinkunjachan: Do we require this shall we put it directly inside the `row.Scan `?
```suggestion

```

internal/reconciler/export_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"context"
+
+	"github.com/openkcm/orbital"
+
+	"github.com/openkcm/krypton/internal/config"
+)
+
+const DefaultMaxPendingReconciles = defaultMaxPendingReconciles
```
jithinkunjachan, line 11: ✅

pkg/api/v1/proto/keys/service.go
```diff
@@ -0,0 +1,67 @@
...
+
+func NewService(keyStore store.Key) *Service {
+	return &Service{keyStore: keyStore}
+}
+
+func (s *Service) AnnounceKey(ctx context.Context, req *AnnounceKeyRequest) (*AnnounceKeyResponse, error) {
+	var parentID *string
+	if req.GetParentId() != "" {
+		p := req.GetParentId()
+		parentID = &p
```
jithinkunjachan: A new feature
```suggestion
		parentID = new(req.GetParentId())
```

jithinkunjachan (commented):
Nicely done

fabenan-f (approved):
lgtm

pkg/store/sql/key.go
```diff
@@ -0,0 +1,90 @@
...
+		key.ManagedBy,
+		labelsJSON,
+		key.State,
+		key.CreatedAt,
+		key.UpdatedAt,
+	)
+	return err
+}
+
+func (ks *KeyStore) GetKeyByID(ctx context.Context, id, tenantID string) (*model.Key, error) {
```
jithinkunjachan, line 49: As we have used the same pattern of Query and Result in other place can we use the same here.
```suggestion
func (ks *KeyStore) GetKeyByID(ctx context.Context, GetKeyQuery..) (GetKeyResult, error) {
```

jithinkunjachan (approved):

PR #138, opened 2026-05-19
feat: add identity package with URI-based entity model

fabenan-f:
Adds the `identity` package defining a `kryptonid://` URI scheme for representing Krypton entities as structured domain + type/name field pairs, with validation and parsing.

### DIFF
--- a/internal/identity/identity.go
+++ b/internal/identity/identity.go
@@ -22,8 +22,8 @@ var (
 	ErrEmptyDomain = errors.New("domain cannot be empty")
 
 	// ErrInvalidCharacters is returned when a domain, field type, or field name
-	// contains characters from invalidChars ('*', '/', or whitespace).
-	ErrInvalidCharacters = errors.New("must not contain '*', '/', or whitespace")
+	// contains characters from invalidChars.
+	ErrInvalidCharacters = errors.New("contains invalid characters")
 
 	// ErrInvalidPath is returned when the path has an incomplete type/name pair.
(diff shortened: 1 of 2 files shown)


internal/identity/identity.go
```diff
@@ -0,0 +1,133 @@
...
+// Field represents a type/name pair in the identity path.
+type Field struct {
+	Type string
+	Name string
+}
+
+// Validate checks that type and name are non-empty and do not contain invalid characters.
+func (f Field) Validate() error {
+	for _, v := range []string{f.Type, f.Name} {
+		if v == "" {
```
apatsap: NP: I think we should either declare all types like this to be consistent with the `const`

```go
type ( 
     T0 foo
     T1 bar
     ....  
)
```

and then all the method

OR

Do  
```
type 
Method
```

But here its a bit of a mix

internal/identity/identity.go
```diff
@@ -0,0 +1,133 @@
...
+		b.WriteByte('/')
+		b.WriteString(f.Name)
+	}
+
+	return URI(b.String())
+}
+
+// Parse parses a kryptonid:// URI into an Identity.
+// Returns an error if the scheme is wrong, domain is empty or invalid,
+// or the path has an odd number of fields (incomplete type/name pair).
```
apatsap, line 99: If we could add a short example URI to the comment it woudl help to better understand why we need an even number of fields

apatsap (changes requested):
nice. just some small nps

apatsap (approved):

jithinkunjachan (approved):

PR #140, opened 2026-05-21
feat: implement AES-256-GCM encrypt/decrypt with secure memory

jithinkunjachan:
Add AES-256-GCM cryptor implementation that performs authenticated encryption and decryption using mlock'd memory for all sensitive data. Includes round-trip tests covering AAD, wrong key, and tampered ciphertext scenarios.

### DIFF
--- a/internal/cryptor/aes256gcm.go
+++ b/internal/cryptor/aes256gcm.go
@@ -1,0 +1,189 @@
+package cryptor
+
+import (
+	"context"
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"errors"
+	"fmt"
+
+	"github.com/openkcm/krypton/internal/securemem"
+)
+
+const (
+	plainTextKey  = "plainText"
+	cipherTextKey = "cipherText"
+)
+
+type AES256GCM struct {
+	info Info
+}
+
+var _ Cryptor = &AES256GCM{}
+
+var ErrAllocatedDataNotFound = errors.New("allocated data not found in vault")
+
+func NewAES256GCM(name string) *AES256GCM {
+	return &AES256GCM{
+		info: Info{
+			Name:                     name,
+			DecryptionSecretRequired: true,
+		},
+	}
+}
+
+// Info returns metadata about the AES256GCM cryptor.
+func (a *AES256GCM) Info() Info {
+	return a.info
+}
+
+// Encrypt encrypts the plaintext using AES-256 in GCM mode with the provided key and AAD.
+func (a *AES256GCM) Encrypt(ctx context.Context, req EncryptRequest) (*EncryptResponse, error) {
+	if err := req.Validate(); err != nil {
+		return nil, err
+	}
+
+	if req.Secret == nil {
+		return nil, fmt.Errorf("missing encryption secret: %w", ErrRequest)
+	}
+
+	secretSize := len(req.Secret.SecureBytes())
+	if secretSize != 32 {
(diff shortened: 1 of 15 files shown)


internal/cryptor/aes256gcm.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+var _ Cryptor = &AES256GCM{}
+
+var ErrAllocatedDataNotFound = errors.New("allocated data not found in vault")
+
+func NewAES256GCM(name string) *AES256GCM {
+	return &AES256GCM{
+		info: Info{
+			Name:                     name,
```
apatsap: This must not be settable by the user. Because this is the field which is used later by the topology/hierarchy.

So we should have something like `AES256GCM` and make this a const

internal/cryptor/aes256gcm.go
```diff
@@ -0,0 +1,189 @@
...
+	return a.info
+}
+
+// Encrypt encrypts the plaintext using AES-256 in GCM mode with the provided key and AAD.
+func (a *AES256GCM) Encrypt(ctx context.Context, req EncryptRequest) (*EncryptResponse, error) {
+	if err := req.Validate(); err != nil {
+		return nil, err
+	}
+
+	if req.Secret == nil {
```
apatsap, line 47: Nice seperation!

internal/cryptor/aes256gcm.go
```diff
@@ -0,0 +1,189 @@
...
+			return fmt.Errorf("failed to allocate secure memory for plaintext: %w", err)
+		}
+
+		// 6. Decrypt and authenticate. Open appends the plaintext into plainBytes'
+		//    backing array (passed as [:0] so it writes from the start).
+		//    Input layout: [nonce || ciphertext || tag]
+		_, err = gcm.Open(
+			plainBytes[:0],
+			req.Ciphertext.SecureBytes()[:nonceSize],
+			req.Ciphertext.SecureBytes()[nonceSize:],
```
apatsap: NP:
can we make it

```go
cipherBytes := req.Ciphertext.SecureBytes()

_, err = gcm.Open(
			plainBytes[:0],
			cypherBytes[:nonceSize],
			cypherBytes[nonceSize:],
```
Its slightly easier to digest

apatsap (changes requested):
Good stuff!!!!

apatsap (approved):

internal/securemem/data_test.go
```diff
@@ -425,3 +426,81 @@ func TestName(t *testing.T) {
...
+		sameUnderLying := func(a, b []byte) bool {
+			if len(a) == 0 || len(b) == 0 {
+				return false
+			}
+			aStart := unsafe.SliceData(a)
+			aEnd := unsafe.Add(unsafe.Pointer(aStart), uintptr(cap(a)-1)*unsafe.Sizeof(a[0]))
+			bStart := unsafe.SliceData(b)
+			bEnd := unsafe.Add(unsafe.Pointer(bStart), uintptr(cap(b)-1)*unsafe.Sizeof(b[0]))
+
+			return uintptr(unsafe.Pointer(aStart)) <= uintptr(bEnd) &&
```
fabenan-f: Could we make this more restrictive by testing for containment instead of just overlap? That is, verify that b fully contains a:

```
return uintptr(unsafe.Pointer(bStart)) <= uintptr(unsafe.Pointer(aStart)) &&
    uintptr(aEnd) <= uintptr(bEnd)
```

fabenan-f (approved):
nice, looks good to me!

fabenan-f (approved):

NamanBalaji (approved):

PR #143, opened 2026-05-26
feat: implemented announce key reconciliation

NamanBalaji:


cmd/agent/main.go
```diff
@@ -94,26 +121,55 @@ func handleErr(err error, msg string) {
...
+		cfg.KryptonRoot.Address.URL = "localhost:" + portOverride
 	}
 
 	return cfg
 }
+
+func setupOperator(ctx context.Context) (*sql.DB, *rpc.Server, *orbital.Operator) {
+	dsn := os.Getenv("AGENT_DATABASE_URL")
+	if dsn == "" {
+		log.Println("AGENT_DATABASE_URL not set, operator disabled")
```
fabenan-f: I know this is only an example agent, but still I'd not allow an agent to run without an operator

integration/helpers_test.go
```diff
@@ -0,0 +1,276 @@
...
+		}
+		select {
+		case <-ctx.Done():
+			t.Fatalf("timed out waiting for job with external_id=%s to reach status %s", externalID, expectedStatus)
+		case <-ticker.C:
+		}
+	}
+}
+
+// awaitKeyExists polls the keys table until a key with the given ID and tenant exists.
```
fabenan-f, line 242: We can make the tests even more integrational if we use the gRPC client for key assertions

internal/agent/handler/announcekey.go
```diff
@@ -0,0 +1,42 @@
+package handler
```
fabenan-f: Imho having `handler.NewAnnounceKey` and `announcekey.NewHandler` is confusing. I think we should consolidate these into a single `handler` package that contains both the job handler (manager side) and the request handler (operator side). This would be beneficial because:

1. Both handlers share common data structures
2. Their interconnection would be immediately visible
3. It would eliminate the current naming ambiguity

internal/agent/handler/announcekey.go
```diff
@@ -0,0 +1,42 @@
...
+		if data.ParentID != "" {
+			parentID = &data.ParentID
+		}
+
+		key := model.NewKey(data.TenantID, data.Name, data.Kind, parentID, data.Target, data.Labels)
+		key.ID = data.KeyID
+		key.State = model.KeyStatePreActivation
+
+		if _, err := keyStore.CreateKey(ctx, store.CreateKeyQuery{Key: key}); err != nil {
+			resp.Fail(fmt.Sprintf("store key: %v", err))
```
fabenan-f: The default case should be to continue, only known terminal errors (like the one in the test case) should fail

internal/keylifecycle/keylifecycle.go
```diff
@@ -26,9 +26,10 @@ var (
 var defaultLifecycle = lifecycle{
 	transitions: map[model.KeyState]map[model.KeyState]struct{}{
 		model.KeyStatePreActivation: {
-			model.KeyStateDestroyed:   {},
-			model.KeyStateActive:      {},
-			model.KeyStateCompromised: {},
+			model.KeyStateDestroyed:      {},
+			model.KeyStateActive:         {},
+			model.KeyStateCompromised:    {},
+			model.KeyStateAnnounceFailed: {},
```
fabenan-f: If we introduce a new state here, we'll deviate from the official NIST lifecycle definition. I'm uncertain whether this is advisable from either a compliance or signaling standpoint

internal/reconciler/handler/announcekey/handler_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/openkcm/krypton/internal/reconciler/handler/announcekey"
+	"github.com/openkcm/krypton/pkg/model"
+	"github.com/openkcm/krypton/pkg/store"
+)
+
+type fakeKeyStore struct {
+	keys          map[string]*model.Key
```
fabenan-f: I'd rather use the real sql store implementation (unhappy paths can still be mocked with wrapper functions)

pkg/api/v1/proto/admin/key_service.go
```diff
@@ -37,18 +45,49 @@ func (s *KeyService) AnnounceKey(ctx context.Context, req *AnnounceKeyRequest) (
...
+
+	data, err := json.Marshal(taskData)
+	if err != nil {
+		return nil, proto.ErrDetailsWithCode(
+			status.New(codes.Internal, "failed to marshal job data"),
+			proto.Code_ERROR_CODE_ABORT,
+		)
+	}
+
+	job := orbital.NewJob(announcekey.JobType, data).WithExternalID(key.ID)
```
fabenan-f: Since `key.ID` is generated internally, a lost response could cause the client to retry, resulting in duplicate jobs performing the same action

fabenan-f (commented):
I like the end-to-end-approach, just a few remarks from my side

internal/agent/handler/announcekey.go
```diff
@@ -0,0 +1,42 @@
...
+			resp.Fail(fmt.Sprintf("unmarshal task data: %v", err))
+			return
+		}
+
+		var parentID *string
+		if data.ParentID != "" {
+			parentID = &data.ParentID
+		}
+
+		key := model.NewKey(data.TenantID, data.Name, data.Kind, parentID, data.Target, data.Labels)
```
jithinkunjachan: ~~Not for this PR, but just a question how the tenants info will be propagated to agent DB as we have DB constraints. We might need to think about  this topic later.~~

internal/keylifecycle/keylifecycle.go
jithinkunjachan: Correct we should have a different processing state 
As per some discussion we should have a field a new field called `KeyProcessingState` and existing `KeyState` should be renamed to `KeyLifeCycleState` 

```golang 
type Key struct {
	ID                 string             `json:"id"`
	Name               string             `json:"name"`
	TenantID           string             `json:"tenant_id"`
	Kind               KeyKind            `json:"kind"`
	ParentID           *string            `json:"parent_id"`
	ManagedBy          string             `json:"managed_by"`
	Labels             Labels             `json:"labels"`
	KeyLifeCycleState  KeyLifeCycleState  `json:"key_lifecycle_state"`
	KeyProcessingState KeyProcessingState `json:"key_processing_state"`
	CreatedAt          clock.UnixNano     `json:"created_at"`
	UpdatedAt          clock.UnixNano     `json:"updated_at"`
}

type KeyProcessingState struct {
	Status string `json:"status"`
	JobID  string `json:"job_id,omitempty"` 
}
```
Here `jobID` shows which JobID is having the lock on , and this will be useful when we do a large key rotation. 
but maybe we can start without and extend it later

Credit to @apatsap as well

jithinkunjachan (commented):
Nice job 👍🏽 , just few things

cmd/agent/main.go
```diff
@@ -65,13 +72,26 @@ func main() {
 	handleErr(err, "failed to create heartbeat worker")
 	go wrkr.Start(ctx)
 
+	// agent-side operator: own database, rpc.Server, handler registration
+	agentDB, rpcServer, operator := setupOperator(ctx)
+	defer agentDB.Close()
+
+	go func() {
+		if err := operator.ListenAndRespond(ctx); err != nil {
```
apatsap, line 81: can we add a log.Info before `ListenAndRespond`

examples/root.config.yaml
apatsap, line 1: can we put this in `/examples/`

examples/agent.config.yaml
apatsap, line 1: can we put this in `/examples/`

internal/handler/announcekey/job.go
```diff
@@ -0,0 +1,90 @@
...
+	return JobType
+}
+
+func (h *JobHandler) ConfirmJob(ctx context.Context, job orbital.Job) (orbital.JobConfirmerResult, error) {
+	var data TaskData
+	if err := json.Unmarshal(job.Data, &data); err != nil {
+		return orbital.CancelJobConfirmer(fmt.Sprintf("invalid job data: %v", err)), nil
+	}
+
+	_, err := h.keyStore.GetKeyByID(ctx, data.KeyID, data.TenantID)
```
apatsap: we need to check that the key state allows the target state using `keylifecycle.ValidateTransition`

internal/handler/announcekey/announcekey.go
```diff
@@ -0,0 +1,31 @@
...
+
+const (
+	JobType  = "announce-key"
+	TaskType = "announce-key"
+)
+
+// TaskData is the payload exchanged between the root job handler and the
+// agent task handler. It is JSON-encoded into the orbital Job/Task data
+// field.
+type TaskData struct {
```
apatsap, line 28: For now its alright. But lets add a comment here that when we add the next handler that certain parts of the data (e.g. Target, Labels, TenantID) might need to be refactored into something like `common.TaskInfo`

internal/handler/announcekey/task.go
```diff
@@ -0,0 +1,90 @@
...
+	return func(ctx context.Context, req orbital.HandlerRequest, resp *orbital.HandlerResponse) {
+		var data TaskData
+		if err := json.Unmarshal(req.TaskData, &data); err != nil {
+			// Terminal: payload is corrupt, retrying won't help.
+			resp.Fail(fmt.Sprintf("unmarshal task data: %v", err))
+			return
+		}
+
+		var parentID *string
+		if data.ParentID != "" {
```
apatsap, line 47: where is checked that, when `data.ParentID` is empty we need to have a root key. Where are we checking that?

internal/handler/announcekey/task.go
```diff
@@ -0,0 +1,90 @@
...
+			parentID = &data.ParentID
+		}
+
+		key := model.NewKey(data.TenantID, data.Name, data.Kind, parentID, data.Target, data.Labels)
+		key.ID = data.KeyID
+		key.LifeCycleState = model.KeyLifeCyclePreActivation
+
+		err := keyStore.CreateKey(ctx, key)
```
apatsap, line 55: we need to make sure `parentID` key exists and can be used to from a lifecycle perspective to announce this key

internal/handler/announcekey/task.go
```diff
@@ -0,0 +1,90 @@
...
+			// Terminal: payload is corrupt, retrying won't help.
+			resp.Fail(fmt.Sprintf("unmarshal task data: %v", err))
+			return
+		}
+
+		var parentID *string
+		if data.ParentID != "" {
+			parentID = &data.ParentID
+		}
+
```
apatsap, line 50: we need to check that data.TenantID exists

pkg/api/v1/proto/admin/key_service.go
```diff
@@ -37,16 +61,97 @@ func (s *KeyService) AnnounceKey(ctx context.Context, req *AnnounceKeyRequest) (
 		req.GetLabels(),
 	)
```
apatsap, line 138: We need to check
* tenantID exists
* Parent exists within the same tenant. 
* If Parent doesn't exist we need to make sure key is root jkey
* If Parent exists we need to check that parent:
   +  is in an allowed key lifecycle state
   + that the keyKind of the new key can be attached to the parent key by the hierarchy definition

pkg/api/v1/proto/admin/key_service.go
```diff
@@ -37,16 +61,97 @@ func (s *KeyService) AnnounceKey(ctx context.Context, req *AnnounceKeyRequest) (
...
 
-	if err := s.keyStore.CreateKey(ctx, key); err != nil {
+	key, err := s.upsertKey(ctx, newKey)
```
apatsap: for the double commit to work we should create the job first and then check in the jobConfirm func whether the key exists and can be activated

pkg/api/v1/proto/admin/key_service.go
```diff
@@ -37,16 +61,97 @@ func (s *KeyService) AnnounceKey(ctx context.Context, req *AnnounceKeyRequest) (
...
+	if err != nil {
+		return nil, err
+	}
+
+	// If a job is already linked, the caller is retrying — return as-is.
+	if key.KeyProcessingState.JobID != "" {
```
apatsap: What if the previous job failed (i.e. it has a job id) and this call's intention is to retry the job

apatsap (changes requested):

pkg/api/v1/proto/admin/key_service.go
NamanBalaji: For now just allowing active state and new key being generated from it's direct parent.

pkg/api/v1/proto/admin/key_service.go
NamanBalaji: Good catch

pkg/api/v1/proto/admin/key_service.go
NamanBalaji: We look up the key by name and if we find a key in failed state we create a new job, we use oldJobID concatenated with the keyID as an externalID. This is effectively our retry job and by using oldJobId we kind of dedup it.

internal/handler/announcekey/job.go
NamanBalaji: imo keylifecycle.ValidateTransition is more for checking state transitions, but in this case we just create a key and for the whole announce operation it will stay in a preactive state so there's effectively no transition taking place here.

internal/handler/announcekey/task.go
NamanBalaji: Will be done as part of a subsequent PR

internal/handler/announcekey/task.go
NamanBalaji: Will be done as part of a subsequent PR

internal/handler/announcekey/task.go
NamanBalaji: Will be done as part of a subsequent PR

apatsap (approved):

internal/handler/announcekey/job.go
```diff
@@ -0,0 +1,105 @@
...
+	var data TaskData
+	if err := json.Unmarshal(job.Data, &data); err != nil {
+		return orbital.CancelJobConfirmer(fmt.Sprintf("invalid job data: %v", err)), nil
+	}
+
+	key, err := h.keyStore.GetKeyByID(ctx, data.KeyID, data.TenantID)
+	if err != nil {
+		if errors.Is(err, store.ErrKeyNotFound) {
+			// ConfirmJob is idempotent and orbital will eventually
+			// time out the job if the key never lands.
```
fabenan-f, line 43: Jobs do not time out in the confirming phase but we can think of something in the future

pkg/api/v1/proto/admin/key_service.go
```diff
@@ -2,51 +2,285 @@ package admin
...
 func (s *KeyService) AnnounceKey(ctx context.Context, req *AnnounceKeyRequest) (*AnnounceKeyResponse, error) {
-	var parentID *string
-	if req.GetParentId() != "" {
-		parentID = new(req.GetParentId())
+	keySpec, vErr := s.validateAnnounceRequest(ctx, req)
+	if vErr != nil {
+		return nil, vErr
+	}
+
+	existing, lookupErr := s.keyStore.GetKeyByName(ctx, store.GetKeyByNameQuery{
```
fabenan-f, line 60: We could be more optimistic and try to create a key first and handle a potential existing key error later, but nothing that needs to be done now

fabenan-f (approved):
The uniqueness constraint on key.Name can detect concurrent jobs when used as an external ID. This might let us remove some of the conditional logic we currently have in the double commit. But the current approach should work as well

jithinkunjachan (approved):
Nicely done ✅

PR #148, opened 2026-05-28
feat: add identity selector

fabenan-f:
Add identity selector with pattern string parsing, validation and identity matching.

### DIFF
--- a/internal/identity/selector.go
+++ b/internal/identity/selector.go
@@ -1,0 +1,187 @@
+package identity
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+var ErrInvalidSelectorPath = errors.New("invalid selector path")
+
+const (
+	// AnyName matches any single name in a field position.
+	AnyName = "*"
+
+	// AnyRemaining matches all remaining fields from its position onward.
+	AnyRemaining = "**"
+)
+
+type (
+	// FieldSelector matches a single Field.
+	FieldSelector struct {
+		Type string
+		Name string
+	}
+
+	// Selector matches identities using a kryptonid:// URI with support for wildcards:
+	//
+	//   - kryptonid://domain/type/name              exact field match, no further fields
+	//   - kryptonid://domain/type/*                 any name for the type, no further fields
+	//   - kryptonid://domain/type/*/othertype/name  any name for first type, exact second field
+	//   - kryptonid://domain/type/**                type must match, any name and any further fields
+	//   - kryptonid://domain/**                     any identity in the domain
+	//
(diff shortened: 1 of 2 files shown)


internal/identity/selector_test.go
```diff
@@ -0,0 +1,652 @@
...
+			fsel:     identity.FieldSelector{Type: "node", Name: identity.AnyName},
+			field:    identity.Field{Type: "node", Name: "anything"},
+			expMatch: true,
+		},
+		{
+			name:     "wildcard name requires type match",
+			fsel:     identity.FieldSelector{Type: "node", Name: identity.AnyName},
+			field:    identity.Field{Type: "user", Name: "anything"},
+			expMatch: false,
+		},
```
jithinkunjachan, line 307: Maybe we can add these 1 test cases as well.
```suggestion
		},		
		{
			name:     "type any remaining matches any field",
			fsel:     identity.FieldSelector{Type: identity.AnyRemaining},
			field:    identity.Field{Type: "node", Name: "root"},
			expMatch: true,
		}

```

internal/identity/selector.go
```diff
@@ -0,0 +1,187 @@
...
+
+	parts := strings.Split(path, "/")
+
+	for i, p := range parts {
+		if p == AnyRemaining && i != len(parts)-1 {
+			return nil, fmt.Errorf("%w: ** must be the last segment", ErrInvalidSelectorPath)
+		}
+	}
+
+	var fields []FieldSelector
```
jithinkunjachan: What about this 
```suggestion
	fields := make([]FieldSelector, 0, (len(parts)+1)/2)
```

jithinkunjachan (approved):
Nicely done

NamanBalaji (approved):

internal/identity/selector_test.go
fabenan-f: ~~I think it is already covered by~~
```go
{
	name: "any remaining matches everything from position",
	selector: identity.Selector{
		Domain: "acme",
		Fields: []identity.FieldSelector{{Type: identity.AnyRemaining}},
	},
	identity: identity.Identity{
		Domain: "acme",
		Fields: []identity.Field{
			{Type: "node", Name: "root"},
			{Type: "service", Name: "kms"},
		},
	},
	expMatch: true,
},
```
I was looking at the wrong test func and added this case

internal/identity/selector.go
fabenan-f: Good point!

jithinkunjachan (approved):
Nice

PR #149, opened 2026-06-01
feat: add AES-256 secret key generation with secure memory

jithinkunjachan:
Add simple secret generator

### DIFF
--- a/internal/cryptor/secretgen.go
+++ b/internal/cryptor/secretgen.go
@@ -45,10 +45,9 @@ func (a *AES256SecretGenerator) GenerateSecret(ctx context.Context) (*GenerateSe
 			return fmt.Errorf("failed to allocate new securemem bytes: %w", err)
 		}
 
-		_, err = rand.Read(b)
-		if err != nil {
-			return fmt.Errorf("failed to generate random bytes: %w", err)
-		}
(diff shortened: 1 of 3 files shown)


internal/cryptor/secretgen.go
```diff
@@ -0,0 +1,72 @@
...
+	resp, err := securemem.Run(ctx, func(ctx context.Context, hReq *securemem.HandlerRequest) error {
+		b, err := hReq.PersistentVault().Reserve(secretKey, 32)
+		if err != nil {
+			return fmt.Errorf("failed to allocate new securemem bytes: %w", err)
+		}
+
+		_, err = rand.Read(b)
+		if err != nil {
+			return fmt.Errorf("failed to generate random bytes: %w", err)
+		}
```
apatsap: Technically rand.Read never returns an error. The signature is chosen like this only to satisfy `io.Reader`

https://pkg.go.dev/crypto/rand#Read

So we could make it shorter and add a comment about the no-error behaviour

apatsap (approved):
Just one small NP. Its your call if you want to change it

fabenan-f (approved):
lgtm

fabenan-f (approved):

PR #150, opened 2026-06-04
feat: add vault interface

fabenan-f:
Introduces the `vault.Vault` interface for secure storage and retrieval of versioned key material (import, export, destroy). Includes `sqlitevault.Unsafe`, an unencrypted SQLite-backed implementation intended for local development and testing only.


internal/vault/unsafe.go
```diff
@@ -0,0 +1,181 @@
+package vault
```
apatsap: can we put the sqllite into its own pkg. Its a bit more structured/contained if we later add more vaults (~10).

```
internal/vault/
       vault.go
       sqlitevault/
           unsafesqllite.go
           ....
```

apatsap (changes requested):
Nice. Just one remark regarding the pkg structure

internal/vault/unsafe.go
fabenan-f: Probably better yes. I just saw the flattened approach in the `cryptor` pkg and followed that pattern but I guess the number of cryptor impl is less

apatsap (approved):

jithinkunjachan (approved):
Nicely done  ✅

PR #157, opened 2026-06-10
feat: add key processor repository

fabenan-f:


integration/helpers_test.go
```diff
@@ -88,17 +88,16 @@ selector_labels:
...
+      name: root-cryptor
+      type: hsm
   K1:
     vault:
       name: root-vault
-      type: in-memory
-      config:
-        prefix: root-kek
+      type: unsafe-sqlite-memory
+    cryptor:
```
apatsap, line 98: Cryptor is very go-ish. 

What about just calling it `crypto` or `cipher`?

internal/cryptor/staticsecret/staticsecret.go
```diff
@@ -12,6 +12,18 @@ import (
 	"github.com/openkcm/krypton/internal/securemem"
 )
 
+const TypeHSM cryptor.Type = "hsm"
```
apatsap, line 15: Lets not call it HSM.

examples/root.config.yaml
```diff
@@ -7,17 +7,16 @@ selector_labels:
...
       name: root-vault
-      type: in-memory
-      config:
-        prefix: root-kek
+      type: unsafe-sqlite-memory
+    cryptor:
+      name: root-cryptor
+      type: aes256gcm
     parent_key_provider:
       agent_name: root
```
apatsap, line 21: I know you didn't add this but this is wrong. Since the root doesn't have `parent_key_provider` but rather K1's cryptor knows how to access  the parent.

Maybe you could just remove it

internal/cryptor/aes256gcm/aes256gcm.go
```diff
@@ -21,6 +21,18 @@ const (
 	cipherTextKey = "cipherText"
 )
 
+const TypeAES256GCM cryptor.Type = "aes256gcm"
+
+// CryptorConfig is the configuration for an AES-256-GCM cryptor.
+type CryptorConfig struct{}
```
apatsap, line 27: Nice

internal/spec/cryptor.go
```diff
@@ -0,0 +1,114 @@
...
+
+var (
+	_ yaml.Unmarshaler = (*CryptorSpec)(nil)
+	_ yaml.Marshaler   = (*CryptorSpec)(nil)
+)
+
+type CryptorBundle struct {
+	Cryptor         cryptor.Cryptor
+	SecretGenerator cryptor.SecretGenerator
+}
```
apatsap, line 28: This is a good idea. But it feels at a wrong place. I think this should live at `cryptor.Bundle`

internal/spec/topology.go
```diff
@@ -40,22 +29,11 @@ type ParentKeyProviderRef struct {
...
-	Type   VaultType   `yaml:"type"` // Vault type (e.g., "open-bao", "aws-kms", "gcp-kms")
-	Config VaultConfig `yaml:"-"`    // Decoded vault config (not from YAML)
-}
-
 // KeyBinding encapsulates all dependencies needed to implement a key kind
 type KeyBinding struct {
-	Vault             VaultSpec             `yaml:"vault"`                         // Storage backend configuration
-	ParentKeyProvider *ParentKeyProviderRef `yaml:"parent_key_provider,omitempty"` // Where to get parent keys for unwrapping
+	Vault             *VaultSpec            `yaml:"vault"`
+	Cryptor           CryptorSpec           `yaml:"cryptor"`
```
apatsap, line 35: Idea: we keep the Krypton terminology inside the struct but only change the yaml flag.

```go	
Cryptor           CryptorSpec           `yaml:"crypto"`
```
```go	
Cryptor           CryptorSpec           `yaml:"cipher"`
```

internal/spec/cryptor.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/openkcm/krypton/internal/cryptor"
+	"github.com/openkcm/krypton/internal/cryptor/aes256gcm"
+	"github.com/openkcm/krypton/internal/cryptor/staticsecret"
+)
+
+type CryptorSpec struct {
+	Name   string         `yaml:"name"`
+	Type   cryptor.Type   `yaml:"type"`
+	Config cryptor.Config `yaml:"-"`
```
apatsap, line 17: I think you've discovered a missing piece.

```go
	Config cryptor.Config `yaml:"-"`
```
This is **very** crucial. Because it tells the cryptor how to initialize.

Example:
Suppose we have 

```yaml
 K3:
          vault:
            name: agent-dek-vault
            ...
          cryptor:
            name: agent-cryptor
            type: static-secret
            config:
                secret:
                    type: k8
                    value: name-of-k8-secret
                // OR
                 secret:
                    type: env
                    value: name-of-env-to-use
               // OR
                  secret:
                    type: unsafe
                    value: this-is the-raw-secret
```

Now, Krypton needs to initialize the `cryptor.StaticSecret` with the config, so we need something like:


```go
in package cryptorprovider

type Provider func(c cryptor.Config)(cryptor.Bundle,error)
```

> Note: We can't add all cryptors inside the `cryptor` package since we'd sooner or later run into cyclic dependencies.


So we'd get
```
    cryptor/
          aes256/
          staticsecret/
          provider/
                cryptorprovider.go
    cryptor.go
```

Inside the `cryptorprovider.go` we'd add all cryptors inside the `init()` function



So all of the `Open` logic would go aways, since each pkg (who has implementation knowledge) knows how to create the `cryptor.Bundle`

Inside the `keyprocessor.NewRepository`

Instead of:
```go
bundle, err := binding.Cryptor.Open(ctx)
		if err != nil {
			return nil, fmt.Errorf("kind %q: open cryptor: %w", kind, err)
		}
```

we'd have
``` 
bundle, err := cryptorprovider.GetBundle(binding.Cryptor.Type)
		if err != nil {
			return nil, fmt.Errorf("kind %q: open cryptor: %w", kind, err)
		}
```


This would mean we get rid of all the implementation logic inside the spec and move it over to a cleaner provider pattern.

> ‼️‼️‼️‼️‼️**NOTE**: we should use the same (conventional) pattern for the vault.

apatsap, line 17: The yaml shouldn't be excluded. I.e. 

```go
type CryptorSpec struct {
...
	Config cryptor.Config `yaml:"config"`
}
```

internal/spec/cryptor.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	if c.Config != nil {
+		if err := c.Config.ValidateCryptorConfig(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *CryptorSpec) Open(ctx context.Context) (CryptorBundle, error) {
```
apatsap, line 42: I think the Open should go away (see long comment) above

internal/spec/cryptor.go
```diff
@@ -0,0 +1,114 @@
...
+
+	*c = CryptorSpec(raw.alias)
+	c.Config = cfg
+	return nil
+}
+
+func newCryptorConfig(t cryptor.Type) (cryptor.Config, error) {
+	switch t {
+	case aes256gcm.TypeAES256GCM:
+		return &aes256gcm.CryptorConfig{}, nil
```
apatsap, line 108: I think we can make `aes256gcm.CryptorConfig` impelment the `yaml.Marshalller` so that it isn't all scattered around

internal/spec/vault.go
```diff
@@ -0,0 +1,95 @@
...
+	"gopkg.in/yaml.v3"
+
+	"github.com/openkcm/krypton/internal/vault"
+	"github.com/openkcm/krypton/internal/vault/sqlitevault"
+)
+
+type VaultSpec struct {
+	Name   string       `yaml:"name"`
+	Type   vault.Type   `yaml:"type"`
+	Config vault.Config `yaml:"-"`
```
apatsap, line 16: I think this should be not excluded:
```go
	Config vault.Config `yaml:"config"`
```

internal/spec/vault.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	if v.Config != nil {
+		if err := v.Config.ValidateVaultConfig(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (v *VaultSpec) Open(ctx context.Context) (vault.Vault, error) {
```
apatsap, line 36: I'd suggest the same provider mechanism as for the `cryptor` package

apatsap (commented):
Very nice. I think we should revisit the provider mechanism though

fabenan-f:
Addressed and merged across multiple PRs

PR #158, opened 2026-06-10
feat: reworked announce key flow and validation

NamanBalaji:
This PR introduces a better flow for key operations starting with announce key and adds a validation package 

- Announce key request is validated and if we are announcing on root or a root key then no job is created, otherwise a job is created
- If key already exists and is in In progress or completed stage we just return the existing key otherwise it's treated as a retry operation 
- On the confirm side we check if we have the latest job ID if not we cancel the job because it's stale and don't modify the key entry 
- If we are confirming the latest job we revalidate because there can be changes in the state and if validation passes we conform the job otherwise we cancel it and mark the key as failed so it can be retried.

Notes: 
- On Agent side we don't need any validations as whatever the root sends is trusted. In future when we implement operations with higher prio like suspend a parent or ancestor or delete a parent etc we traverse down the subtree cancel all the in progress job for child keys and then perform our operation. So root acts as the orchestrator and agents just accept. 
- For now announce is only possible when the parent is a direct parent of the key that's being announced and parent needs to be active, this can be changed in the future.


integration/announce_key_test.go
```diff
@@ -123,4 +126,52 @@ func TestAnnounceKey(t *testing.T) {
 		assert.Equal(t, first.GetKey().GetId(), second.GetKey().GetId())
 		assert.Equal(t, first.GetKey().GetKeyProcessingState().GetJobId(), second.GetKey().GetKeyProcessingState().GetJobId())
 	})
```
jithinkunjachan, line 152: Shall we have a integration test for this scenario where we announce key for a root agent

```go
		resp, err := keyCli.AnnounceKey(ctx, &keypb.AnnounceKeyRequest{
			TenantId:   tenantID,
			Kind:       "K1",
			Name:       keyName,
			ParentId:   parentID,
			TargetName: "root",
			Labels:     map[string]string{"cloud": "aws"},
		})
```

pkg/api/v1/proto/admin/keys/setup_test.go
```diff
@@ -0,0 +1,222 @@
...
+			{Kind: "K1", Role: spec.KeyRoleKek, Algorithm: cryptor.KeyAlgorithmAES256},
+			{Kind: "K2", Role: spec.KeyRoleTek, Algorithm: cryptor.KeyAlgorithmAES256},
+			{Kind: "K3", Role: spec.KeyRoleDek, Algorithm: cryptor.KeyAlgorithmAES256},
+		},
+	}
+}
+
+// defaultTestTopology pairs with defaultTestHierarchy and exposes two
+// segments: "root" (K0..K0) and "agent" (K1..K3).
+func defaultTestTopology() spec.Topology {
```
jithinkunjachan, line 85: In the integration test we don't have root segment but here we have it.  IMHO If config are wrong we might have some problems so lets be consistent. 
Cause now if we remove the root from the topology 
```go
			{Name: testRootName, Segment: spec.HierarchySegment{StartKind: "K0", EndKind: "K0"}},
```
as we make it similar to the integration test config the test fails for the validation for announcekey(**key_service_test.go:76**).

https://github.com/openkcm/krypton/blob/8c7ee07948e67d5452150e6e467342fdcbd7c9e3/integration/helpers_test.go#L127-L149

jithinkunjachan (commented):
Nice work

pkg/api/v1/proto/admin/keys/key_service.go
```diff
@@ -0,0 +1,257 @@
...
+			return nil, proto.ErrDetailsWithCode(
+				status.New(codes.Internal, fmt.Sprintf("failed to create key, err : %v", err)),
+				proto.Code_ERROR_CODE_RETRY,
+			)
+		}
+
+		return &AnnounceKeyResponse{Key: KeyToProto(newKey)}, nil
+	}
+
+	// key exssist but our announce data doesn't match
```
apatsap: typo `exssist`

pkg/api/v1/proto/admin/keys/key_service.go
```diff
@@ -0,0 +1,257 @@
...
+	if !existing.Equals(&newKey) {
```
apatsap: NP: I stumbled upon the `Equals` since it only checks for certain fields (which makes sense). However, I'd call it `IsSame` or `Same`

apatsap: We end up here independently of whether or not the key exists. So `existing` can be nil. You can either check here for nil or update the `Equals` method

```go
func (k *Key) Equals(other *Key) bool {
    if k == nil {
         return false
    }
...

}
```

api-specs/v1/proto/admin/keys/key.proto
```diff
@@ -1,9 +1,9 @@
...
 
 
-package krypton.v1.admin;
+package krypton.v1.admin.keys;
 
-option go_package = "github.com/openkcm/krypton/pkg/api/v1/proto/admin";
+option go_package = "github.com/openkcm/krypton/pkg/api/v1/proto/admin/keys";
 
 service KeyService {
   rpc AnnounceKey(AnnounceKeyRequest) returns (AnnounceKeyResponse);
```
jithinkunjachan, line 9: Currently `AnnounceKeyRequest` don't have ` --target-selector ` field will this be added in subsequent PR Or we go for lean approach for the time being.

pkg/api/v1/proto/admin/keys/key_service.go
```diff
@@ -0,0 +1,257 @@
...
+		return nil, proto.ErrDetailsWithCode(
+			status.New(codes.FailedPrecondition, conflictingKeyErrMsg),
+			proto.Code_ERROR_CODE_ABORT,
+		)
+	}
+
+	// idempotent if key is was already announced successfully or still being processed
+	if existing.KeyProcessingState.Status == model.KeyProcessingInProgress || existing.KeyProcessingState.Status == model.KeyProcessingCompleted {
```
apatsap: NP for better comprehensibility I'd add a method

```go
existing.KeyProcessingState.Status.IsOneOf(
           model.KeyProcessingInProgress,
           model.KeyProcessingCompleted,
)
```

pkg/api/v1/proto/admin/keys/key_service.go
```diff
@@ -0,0 +1,257 @@
...
+		TenantID: req.GetTenantId(),
+		Name:     req.GetName(),
+	})
+	if err != nil && !errors.Is(err, store.ErrKeyNotFound) {
+		return nil, proto.ErrDetailsWithCode(
+			status.New(codes.Internal, fmt.Sprintf("failed to look up key: %v", err)),
+			proto.Code_ERROR_CODE_RETRY,
+		)
+	}
+
```
apatsap, line 75: From here on I there is a lot of stuff going on, and I suggest to simplify it

If no key exists:
   createKey k (in mem)
   prepareJob for k
   store k in db
if key k exists 
    check if k is in processing state failed
    if yes
        prepareJob for k 

I've added a comment to the confirmer, that it only allows pending/failed states of key 

I think that would simplify a lot and can also be pretty much used for the other lifecylce states

apatsap (changes requested):

api-specs/v1/proto/admin/keys/key.proto
NamanBalaji: Yes I was unsure how that will work currently we are using the target name for that. Ideally if we use something like labels we would need to figure out which labels match which agents and then find the agent where the announce will be valid, and then resolve the tasks. This can be done in a future ticket.

pkg/api/v1/proto/admin/keys/setup_test.go
NamanBalaji: Good catch, I fixed it.

apatsap (approved):

internal/handler/announcekey/job.go
```diff
@@ -12,47 +12,63 @@ import (
...
 	if err != nil {
 		if errors.Is(err, store.ErrKeyNotFound) {
-			// ConfirmJob is idempotent and orbital will eventually
-			// time out the job if the key never lands.
 			return orbital.ContinueJobConfirmer(), nil
 		}
 		return nil, fmt.Errorf("confirm job: %w", err)
 	}
 
+	if key.KeyProcessingState.JobID != job.ID.String() {
```
jithinkunjachan, line 49: good check

jithinkunjachan (approved):
Nice Job

PR #159, opened 2026-06-11
feat: add secret and cryptor provider

fabenan-f:
Add a provider layer for cryptors and an independent secret-loading package so that cryptor bundles and their key material can be resolved from YAML configuration at runtime.

### DIFF
--- a/internal/cryptor/cryptorprovider/provider.go
+++ b/internal/cryptor/cryptorprovider/provider.go
@@ -19,6 +19,7 @@ var ErrCryptorNameEmpty = errors.New("cryptor name is empty")
 type Spec struct {
 	Name   string         `yaml:"name"`
 	Type   cryptor.Type   `yaml:"type"`
+	// Config is marshaled/unmarshaled via custom MarshalYAML/UnmarshalYAML, not by the default YAML codec.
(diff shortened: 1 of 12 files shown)


internal/cryptor/cryptorprovider/provider.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+	return nil
+}
+
+// GetBundle constructs a ready-to-use cryptor Bundle from the given Spec.
+func GetBundle(ctx context.Context, spec Spec) (cryptor.Bundle, error) {
+	switch c := spec.Config.(type) {
+	case *aes256gcm.Config:
+		_ = c // no config fields to use
```
apatsap: why do we need this?

apatsap (approved):
Just one small thing. Not approval relevant. 

Nice that you applied the provider pattern also to the secret pkg

internal/cryptor/cryptorprovider/provider.go
fabenan-f: temporary step during development that is removed now

jithinkunjachan (approved):
Nice Job

PR #162, opened 2026-06-16
feat: add ListKeys with cursor-based pagination and filtering

jithinkunjachan:
Implement ListKeys on the Key store interface with support for:
- Cursor-based pagination using base64-encoded (created_at, id) tuples
- Configurable sort order (ascending/descending by created_at)
- Filtering by kind, lifecycle state, managing agent, name (ILIKE substring), and JSONB label containment
- Default page size of 50 when no limit is specified Also enriches the test key hierarchy (8 → 10 keys) with distinct lifecycle states, managing agents, and labels to support targeted filter assertions.

### DIFF
--- a/pkg/store/sql/cursor_test.go
+++ b/pkg/store/sql/cursor_test.go
@@ -1,0 +1,92 @@
+package sql_test
+
+import (
+	"encoding/base64"
+	"regexp"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/openkcm/krypton/internal/clock"
+	"github.com/openkcm/krypton/pkg/store/sql"
+)
+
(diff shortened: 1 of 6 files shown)


pkg/store/sql/cursor_test.go
```diff
@@ -0,0 +1,92 @@
...
+		},
+		{
+			name:  "empty string",
+			input: "",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := sql.DecodeCursor(tt.input)
+			require.Error(t, err)
```
fabenan-f: `assert.Error(t, err)` in this case?

fabenan-f (approved):
nice job! (just one tiny comment that also can be ignored)

NamanBalaji (approved):
Nice

fabenan-f (approved):

PR #164, opened 2026-06-17
feat: add ListKeys RPC with cursor-based pagination and filtering

jithinkunjachan:
Implements the ListKeys endpoint supporting filters by kind, name, lifecycle state, managed_by, and labels with configurable sort order.


fabenan-f (approved):
good stuff

NamanBalaji (approved):

PR #166, opened 2026-06-19
feat: add get keys CLI command with filtering and pagination

jithinkunjachan:
## Summary
- Add `kr get keys` CLI command that lists keys with filtering and cursor-based pagination
- Make `--tenant-id` optional on `parent-keys`, `descendant-keys`, and `keys` commands by falling back to the selected tenant from the state store


NamanBalaji (approved):
Nice!

fabenan-f (approved):
looks good

PR #168, opened 2026-06-19
feat: add key processor

fabenan-f:
Adds `keyprocessor.Processor` which manages secret lifecycle (create, wrap, unwrap, delete) within a composable key hierarchy, handling layered encryption/decryption and secure memory cleanup.

### DIFF
--- a/internal/keyprocessor/export_test.go
+++ b/internal/keyprocessor/export_test.go
@@ -9,14 +9,13 @@ import (
 	"github.com/openkcm/krypton/pkg/model"
 )
 
-func NewProcessor(name string, v vault.Vault, c cryptor.Cryptor, sg cryptor.SecretGenerator, internal *Processor, parent *Processor) *Processor {
+func NewProcessor(generator cryptor.SecretGenerator, wrapper, sealer cryptor.Cryptor, v vault.Vault, parent *Processor) *Processor {
 	return &Processor{
-		name:            name,
-		vault:           v,
-		cryptor:         c,
-		secretGenerator: sg,
-		internal:        internal,
-		parent:          parent,
+		generator: generator,
+		wrapper:   wrapper,
+		sealer:    sealer,
+		vault:     v,
+		parent:    parent,
 	}
 }
 
--- a/internal/keyprocessor/export_test.go
+++ b/internal/keyprocessor/export_test.go
@@ -1,0 +1,25 @@
+package keyprocessor
+
+import (
+	"context"
+
+	"github.com/openkcm/krypton/internal/cryptor"
+	"github.com/openkcm/krypton/internal/securemem"
+	"github.com/openkcm/krypton/internal/vault"
+	"github.com/openkcm/krypton/pkg/model"
+)
+
+func NewProcessor(name string, v vault.Vault, c cryptor.Cryptor, sg cryptor.SecretGenerator, internal *Processor, parent *Processor) *Processor {
+	return &Processor{
+		name:            name,
+		vault:           v,
+		cryptor:         c,
+		secretGenerator: sg,
+		internal:        internal,
+		parent:          parent,
+	}
+}
+
+func (p *Processor) ResolveSecret(ctx context.Context, keyChain []model.Key) (*securemem.Data, error) {
+	return p.resolveSecret(ctx, keyChain)
+}
--- a/internal/keyprocessor/processor.go
+++ b/internal/keyprocessor/processor.go
@@ -3,6 +3,7 @@ package keyprocessor
 import (
 	"context"
 	"errors"
+	"log/slog"
 
 	"github.com/openkcm/krypton/internal/cryptor"
 	"github.com/openkcm/krypton/internal/securemem"
@@ -10,8 +11,15 @@ import (
 	"github.com/openkcm/krypton/pkg/model"
 )
 
-// ErrEmptyKeyChain is returned when an operation receives an empty key chain.
-var ErrEmptyKeyChain = errors.New("empty key chain")
+const persistSecName = "persistedSec"
+
+var (
+	// ErrEmptyKeyChain is returned when an operation receives an empty key chain.
+	ErrEmptyKeyChain = errors.New("empty key chain")
+	// ErrSecNotPersisted is returned when a handler completes successfully
+	// but the secret is missing from the persistent vault, indicating a bug
+	ErrSecNotPersisted = errors.New("persisted secret missing from handler response")
+)
 
 // Processor creates, wraps, unwraps, and deletes secrets within a key hierarchy.
 // It destroys all intermediate secrets that pass through its operations.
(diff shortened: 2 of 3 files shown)


internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+		if err := destroyAll(sec); err != nil {
+			return CreateSecretResponse{}, err
+		}
+		sec = resp.WrappedSecret
+	}
+
+	_, err = p.vault.ImportKey(ctx, vault.ImportKeyRequest{
+		TenantID:    key.TenantID,
+		KeyID:       p.keyID(key.ID),
+		KeyVersion:  1,
```
fabenan-f: The `cryptor` and `vault` interface already support/require key versions but `model.Key` does not, for now I'd only assume one version and revisit later

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+	}
+	return errors.Join(errs...)
+}
+
+func toCryptorSecret(data *securemem.Data) *cryptor.Secret {
+	if data == nil {
+		return nil
+	}
+	return &cryptor.Secret{
```
fabenan-f, line 364: The `vault` interface does not use `cryptor.Secret`, so we cannot store the `Algorithm`. If later added, we should change it here as well but since we only support `AES256GCM` currently I think it is fine

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	return sec, nil
+}
+
+func (p *Processor) keyID(id string) string {
+	if p.name == "" {
+		return id
+	}
+	return p.name + "/" + id
```
jithinkunjachan: I think as discussed shall we also add tenantID here ?

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+		KeyID:       p.keyID(key.ID),
+		KeyVersion:  1,
+		KeyMaterial: sec,
+		AAD:         req.AAD,
+	})
+	if err != nil {
+		return CreateSecretResponse{}, errors.Join(err, destroyAll(genResp.Secret, sec))
+	}
+
+	if err := destroyAll(genResp.Secret, sec); err != nil {
```
jithinkunjachan:
```suggestion
	return CreateSecretResponse{}, destroyAll(genResp.Secret, sec)
```

or we could something like this 
```go
_, err = p.vault.ImportKey(ctx, vault.ImportKeyRequest{
    TenantID:    key.TenantID,
    KeyID:       p.keyID(key.ID),
    KeyVersion:  1,
    KeyMaterial: sec,
    AAD:         req.AAD,
})
return CreateSecretResponse{}, errors.Join(err, destroyAll(genResp.Secret, sec))
```

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+		KeyVersion: 1,
+		Secret:     toCryptorSecret(sec),
+		Plaintext:  req.Secret,
+		AAD:        req.AAD,
+	})
+	if err != nil {
+		return WrapSecretResponse{}, errors.Join(err, destroyAll(sec))
+	}
+
+	if err := destroyAll(sec); err != nil {
```
jithinkunjachan: IMHO please also try to  destroy `encResp` as this is a `securemem.Data`

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+		Secret:     toCryptorSecret(sec),
+		Ciphertext: req.WrappedSecret,
+		AAD:        req.AAD,
+	})
+	if err != nil {
+		return UnwrapSecretResponse{}, errors.Join(err, destroyAll(sec))
+	}
+
+	if err := destroyAll(sec); err != nil {
+		return UnwrapSecretResponse{}, err
```
jithinkunjachan, line 188: IMHO please also try to destroy `decResp` as this is a `securemem.Data`

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+	if err != nil {
+		return DeleteSecretResponse{}, err
+	}
+
+	if p.internal != nil {
+		_, err = p.internal.DeleteSecret(ctx, DeleteSecretRequest{
+			Key: model.Key{TenantID: req.Key.TenantID, ID: req.Key.ID},
+		})
+		if err != nil {
+			return DeleteSecretResponse{}, err
```
jithinkunjachan: IMHO we could avoid if and directly return
```go
return DeleteSecretResponse{}, err
```

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+	sec := exported.KeyMaterial
+	aad := exported.AAD
+
+	if p.parent != nil {
+		resp, err := p.parent.UnwrapSecret(ctx, UnwrapSecretRequest{
+			KeyChain:      keyChain[:len(keyChain)-1],
+			WrappedSecret: sec,
+			AAD:           aad,
+		})
+		if err != nil {
```
jithinkunjachan: shall we combine everything like this ?
```go
if err = errors.Join(err, destroyAll(sec)); err != nil {
    return nil, err
}
```

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+		sec = resp.Secret
+	}
+
+	if p.internal != nil {
+		resp, err := p.internal.UnwrapSecret(ctx, UnwrapSecretRequest{
+			KeyChain:      []model.Key{{TenantID: key.TenantID, ID: key.ID}},
+			WrappedSecret: sec,
+			AAD:           aad,
+		})
+		if err != nil {
```
jithinkunjachan: same here 
```go
if err = errors.Join(err, destroyAll(sec)); err != nil {
    return nil, err
}
```

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+			return CreateSecretResponse{}, errors.Join(err, destroyAll(genResp.Secret))
+		}
+
+		resp, err := p.internal.WrapSecret(ctx, WrapSecretRequest{
+			KeyChain: []model.Key{{TenantID: key.TenantID, ID: key.ID}},
+			Secret:   sec,
+			AAD:      req.AAD,
+		})
+		if err != nil {
```
jithinkunjachan:
```go
if err = errors.Join(err, destroyAll(sec)); err != nil {
			return CreateSecretResponse{}, err
		}
```

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+		resp, err := p.parent.WrapSecret(ctx, WrapSecretRequest{
+			KeyChain: req.KeyChain[:len(req.KeyChain)-1],
+			Secret:   sec,
+			AAD:      req.AAD,
+		})
+		if err != nil {
+			return CreateSecretResponse{}, errors.Join(err, destroyAll(genResp.Secret, sec))
+		}
+		if err := destroyAll(sec); err != nil {
+			return CreateSecretResponse{}, err
```
jithinkunjachan:
```suggestion
		CreateSecretResponse{}, errors.Join(err,destroyAll(genResp.Secret))
```
Shall we destroy maybe we can combine all this in one `destroyAll`

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+
+	genResp, err := p.secretGenerator.GenerateSecret(ctx)
+	if err != nil {
+		return CreateSecretResponse{}, err
+	}
+
+	sec := genResp.Secret
+
+	if p.internal != nil {
+		_, err := p.internal.CreateSecret(ctx, CreateSecretRequest{
```
jithinkunjachan: Suppose we have an internal `processor` 
and now we `CreateSecret` for this particular internal secret with `TenantID` and `KeyID` and this will call the `CreateSecret` recursively, so it basically create a secret and try to store it.
So now the internal secret is created and stored but when go back for some reason if this `ImportKey` fails cause of some network issue we dont have a rollback of the internal secret, or is it kind of ok to store it internal secret.

jithinkunjachan (commented):
Very good job 🚀

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+type DeleteSecretRequest struct {
+	Key model.Key
+}
+
+// DeleteSecretResponse is the output of DeleteSecret.
+type DeleteSecretResponse struct{}
+
+// CreateSecret generates a secret, wraps it through the internal and parent chain, and stores it in the vault.
+// It is a no-op when the cryptor manages its own decryption key regardless of internal or parent configuration.
+func (p *Processor) CreateSecret(ctx context.Context, req CreateSecretRequest) (CreateSecretResponse, error) {
```
NamanBalaji, line 77: We need to make this work atomically or have some kind of rollback mech or state machine because we can have failures after secret generation that will leave us with orphaned key material.

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+// DeleteSecret removes the processor's secret and its internal processor's secret from the vault.
+// It is a no-op when the cryptor manages its own decryption key regardless of internal or parent configuration.
+func (p *Processor) DeleteSecret(ctx context.Context, req DeleteSecretRequest) (DeleteSecretResponse, error) {
+	if !p.cryptor.Info().DecryptionSecretRequired {
+		return DeleteSecretResponse{}, nil
+	}
+
+	_, err := p.vault.DestroyKey(ctx, vault.DestroyKeyRequest{
```
NamanBalaji, line 204: We destroy this before the internal key, if the destruction of internal fails it will leave it in an orphaned state. IMO we also need to make this work as if it was atomic.

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,323 @@
...
+
+		resp, err := p.internal.WrapSecret(ctx, WrapSecretRequest{
+			KeyChain: []model.Key{{TenantID: key.TenantID, ID: key.ID}},
+			Secret:   sec,
+			AAD:      req.AAD,
+		})
+		if err != nil {
+			return CreateSecretResponse{}, errors.Join(err, destroyAll(genResp.Secret))
+		}
+		if err := destroyAll(sec); err != nil {
```
NamanBalaji: If the destroy fails we will never reassign to `sec` and the newly generated secret leaks.

NamanBalaji (commented):
Nice work

internal/keyprocessor/processor.go
fabenan-f: Thanks for noticing all the early returns! That saved a couple of lines

internal/keyprocessor/processor.go
fabenan-f: You mean `p.name + "/" + tenantID "/" + id`? We can do this but at every place where we provide the prefixed key ID, we pass the tenant ID as well (vault, cryptor)

internal/keyprocessor/processor.go
fabenan-f: I completely agree. Thinking about it more, I don't think we need to autoprovision internal keys here, so I removed it. This can be handled separately via a dedicated call to the internal key processor. This might be a good use case for orbital's job groups with one job per key processor where the terminal func for failed job groups handles the rollback.

I'd suggest we defer this topic to a later story

internal/keyprocessor/processor.go
fabenan-f: Nice catch!

jithinkunjachan (approved):

internal/keyprocessor/export_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"context"
+
+	"github.com/openkcm/krypton/internal/cryptor"
+	"github.com/openkcm/krypton/internal/securemem"
+	"github.com/openkcm/krypton/internal/vault"
+	"github.com/openkcm/krypton/pkg/model"
+)
+
+func NewProcessor(name string, v vault.Vault, c cryptor.Cryptor, sg cryptor.SecretGenerator, internal *Processor, parent *Processor) *Processor {
```
apatsap: NP slightly more expressive due to less chars
```go
func NewProcessor(name string, v vault.Vault, c cryptor.Cryptor, sg cryptor.SecretGenerator, internal, parent *Processor) *Processor 
```

internal/keyprocessor/processor.go
apatsap: I think tenantID is a good idea, because then we have a real tenant separation

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,282 @@
...
+		if err := errors.Join(err, destroySec(sec)); err != nil {
+			return nil, errors.Join(err, destroySec(resp.Secret))
+		}
+		sec = resp.Secret
+	}
+
+	return sec, nil
+}
+
+func (p *Processor) keyID(id string) string {
```
apatsap: NP
lets call it `vaultKeyID`

apatsap (changes requested):

internal/keyprocessor/processor.go
fabenan-f: I removed this explicit namespace by processor for this first iteration to keep the scope cleaner, we can add it later

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,306 @@
...
+type DeleteSecretRequest struct {
+	Key model.Key
+}
+
+// DeleteSecretResponse is the output of DeleteSecret.
+type DeleteSecretResponse struct{}
+
+// CreateSecret generates a secret, seals it, wraps it through the parent chain, and stores it in the vault.
+// It is a no-op when the wrapper manages its own decryption key.
+func (p *Processor) CreateSecret(ctx context.Context, req CreateSecretRequest) (CreateSecretResponse, error) {
```
apatsap, line 77: I think we are still not utilizing securemem fully. Our goal should be that a raw secret lives for as little as possible in memory and it can't be dumped during that period.

This is an example how the securemem should be fully utilized

```go
func (p *Processor) CreateSecret(ctx context.Context, req CreateSecretRequest) (CreateSecretResponse, error) {
    // Wrap the secret creation into a securemem handler so that every allocation is cleaned up and destroyed in a timely manner
	_, err := securemem.Run(ctx, func(ctx context.Context, sreq *securemem.HandlerRequest) error {
		if !p.wrapper.Info().DecryptionSecretRequired {
			return nil
		}

		key, err := lastKey(req.KeyChain)
		if err != nil {
			return err
		}

		genResp, err := p.generator.GenerateSecret(ctx)
		if err != nil {
			return err
		}
		defer genResp.Secret.Destroy()

		sealSec, err := p.seal(ctx, key, genResp.Secret, req.AAD)
		if err != nil {
			return err
		}
		defer sealSec.Destroy()

		wrapSec, err := p.parentWrap(ctx, req.KeyChain, sealSec, req.AAD)
		if err != nil {
			return err
		}
		defer wrapSec.Destroy()

		_, err = p.vault.ImportKey(ctx, vault.ImportKeyRequest{
			TenantID:    key.TenantID,
			KeyID:       key.ID,
			KeyVersion:  1,
			KeyMaterial: wrapSec,
			AAD:         req.AAD,
		})
		return err
	})

	return CreateSecretResponse{}, err
}
```

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,306 @@
...
+
+	sec, err := p.seal(ctx, key, genResp.Secret, req.AAD)
+	if err != nil {
+		return CreateSecretResponse{}, err
+	}
+
+	sec, err = p.parentWrap(ctx, req.KeyChain, sec, req.AAD)
+	if err != nil {
+		return CreateSecretResponse{}, err
+	}
```
apatsap: Generally speaking we shouldn't do that. Because now the first sec is only destroyed when the GC runs eventually (see comment above)

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,306 @@
...
+		KeyVersion:  1,
+		KeyMaterial: sec,
+		AAD:         req.AAD,
+	})
+	return CreateSecretResponse{}, errors.Join(err, destroySec(sec))
+}
+
+// WrapSecret resolves the wrapping secret from the key chain and encrypts the given secret.
+// When the wrapper manages its own decryption key, it encrypts directly without resolving.
+func (p *Processor) WrapSecret(ctx context.Context, req WrapSecretRequest) (WrapSecretResponse, error) {
```
apatsap, line 121: We should apply `securemem.Run` here is as well

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,306 @@
...
+	sec, err := p.resolveSecret(ctx, req.KeyChain)
```
apatsap: The sec must be `defer sec.Destroy` and then we can remove `errors.Join(err, destroySec(sec))`

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	return WrapSecretResponse{
+		WrappedSecret: encResp.Ciphertext,
+	}, nil
+}
+
+// UnwrapSecret resolves the wrapping secret from the key chain and decrypts the given wrapped secret.
+// When the wrapper manages its own decryption key, it decrypts directly without resolving.
+func (p *Processor) UnwrapSecret(ctx context.Context, req UnwrapSecretRequest) (UnwrapSecretResponse, error) {
```
apatsap, line 160: We should apply `securemem.Run` here is as well

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,306 @@
...
+	sec, err := p.resolveSecret(ctx, req.KeyChain)
```
apatsap: The sec must be `defer sec.Destroy()` and then we can remove `errors.Join(err, destroySec(sec))`

internal/keyprocessor/processor.go
```diff
@@ -0,0 +1,306 @@
...
+	})
+	if err := errors.Join(err, destroySec(sec)); err != nil {
+		if decResp != nil {
+			return UnwrapSecretResponse{}, errors.Join(err, destroySec(decResp.Plaintext))
+		}
+		return UnwrapSecretResponse{}, err
+	}
+
+	return UnwrapSecretResponse{
+		Secret: decResp.Plaintext,
```
apatsap: Since we should wrap this inside `securemem.Run` this might lead to problems because of 
```
To reduce lifetimes of allocations and avoid unexpected performance issues, if a function invoked by Do needs to yield a result that shouldn't be erased, it should do so by copying the result into an allocation created by the caller.
```
https://pkg.go.dev/runtime/secret#Do

so we need to call the new memvault import function

```go
req.PersistentVault().Import("result" , decResp)
...

And export it from the securemem.Run repsonse
```

apatsap (changes requested):

internal/keyprocessor/processor.go
fabenan-f: `securemem.Run(...)` is now applied

apatsap (approved):

jithinkunjachan (approved):

PR #173, opened 2026-07-01
feat: copying over KMIP package

NamanBalaji:


nnicora (approved):

jithinkunjachan (approved):

apatsap (approved):

internal/kmip/get_test.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+		sk, ok := resp.Object.(*kmip.SymmetricKey)
+		if !ok {
+			t.Fatalf("Object type = %T", resp.Object)
+		}
+		if sk.KeyBlock.KeyFormatType != kmip.KeyFormatTypeRaw {
+			t.Fatalf("KeyFormatType = %v", sk.KeyBlock.KeyFormatType)
+		}
+		if sk.KeyBlock.CryptographicAlgorithm != kmip.CryptographicAlgorithmAES {
+			t.Fatalf("Algorithm = %v", sk.KeyBlock.CryptographicAlgorithm)
```
jithinkunjachan: please use assert

internal/kmip/get.go
```diff
@@ -0,0 +1,53 @@
...
+	if err := authorizeTenant(ctx, tenantID); err != nil {
+		return nil, toKMIPError(err)
+	}
+
+	dek, err := h.keyManager.GetDEK(ctx, tenantID, keyID)
+	if err != nil {
+		return nil, toKMIPError(err)
+	}
+
+	material := make([]byte, 0, dek.LengthBits/8)
```
jithinkunjachan: we leak the material here as this data bytes is leaving locked memory.
I know this is being cleaned up by the `teminateHook`

internal/kmip/get_attributes.go
```diff
@@ -0,0 +1,55 @@
...
+
+	var attrs []kmip.Attribute
+	if len(req.AttributeName) == 0 {
+		attrs = all
+	} else {
+		requested := make(map[kmip.AttributeName]struct{}, len(req.AttributeName))
+		for _, n := range req.AttributeName {
+			requested[n] = struct{}{}
+		}
+		attrs = make([]kmip.Attribute, 0, len(all))
```
jithinkunjachan: IMHO we are allocating for `all` shall we do it for `len(requested)`

internal/kmip/get_attributes.go
```diff
@@ -0,0 +1,55 @@
...
+	if err := authorizeTenant(ctx, tenantID); err != nil {
+		return nil, toKMIPError(err)
+	}
+
+	info, err := h.keyManager.GetKeyInfo(ctx, tenantID, keyID)
+	if err != nil {
+		return nil, toKMIPError(err)
+	}
+
+	all := []kmip.Attribute{
```
jithinkunjachan: is the ordering of the attributes important otherwise we can convert to a map here and avoid some for loops down, or use `switch` to build it from `req`

internal/kmip/identifier_test.go
```diff
@@ -0,0 +1,43 @@
...
+		{"", "", "", ErrInvalidKeyIdentifier},
+		{":", "", "", ErrInvalidKeyIdentifier},
+	}
+	for _, tt := range tests {
+		t.Run(tt.in, func(t *testing.T) {
+			t.Parallel()
+			gotTenant, gotKey, err := parseKeyIdentifier(tt.in)
+			if tt.wantErr != nil {
+				if !errors.Is(err, tt.wantErr) {
+					t.Fatalf("err = %v, want %v", err, tt.wantErr)
```
jithinkunjachan: IMHO assert

internal/kmip/errors.go
```diff
@@ -0,0 +1,42 @@
...
+// kmipAlgorithm translates the internal Algorithm enum into the kmip-go
+// wire-format constant. It defaults to AES for AlgorithmUnknown so a
+// misconfigured seed still produces a structurally valid response; callers
+// that care should validate before serving.
+func kmipAlgorithm(a Algorithm) kmip.CryptographicAlgorithm {
+	switch a {
+	case AlgorithmAES:
+		return kmip.CryptographicAlgorithmAES
+	default:
+		return kmip.CryptographicAlgorithmAES
```
jithinkunjachan: if somebody forgot to configure the new algorithm then we we sent the the wrong KMIP algorithm would it be better if we fail it ?

internal/kmip/memkeymanager.go
```diff
@@ -0,0 +1,132 @@
...
+	data       *securemem.Data
+	algorithm  Algorithm
+	lengthBits int32
+	state      model.KeyLifeCycleState
+}
+
+// NewMemKeyManager returns a KeyManager seeded with the given DEKs. Each
+// seed's Material is copied into a locked secure-memory region; the caller
+// may zero its own copy after this call returns.
+func NewMemKeyManager(seeds ...SeedDEK) (KeyManager, error) {
```
jithinkunjachan, line 40: how this is will be refreshed from the main if we want to add some more deks, or is it a mock for now?

internal/kmip/memkeymanager_test.go
```diff
@@ -0,0 +1,118 @@
+package kmip
```
jithinkunjachan: package

internal/kmip/tls_test.go
```diff
@@ -0,0 +1,86 @@
+package kmip
```
jithinkunjachan: please have a look into the test `package` and also the `assert`

jithinkunjachan (commented):
Nice work

internal/kmip/wipe.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/openkcm/krypton/internal/securemem"
+)
+
+// wipeRegistry holds byte slices that must be zeroed when a KMIP connection
+// terminates. Each connection has its own registry stored on the connection
+// context; the ConnectHook attaches it and the TerminateHook drains it.
+type wipeRegistry struct {
+	mu      sync.Mutex
+	buffers [][]byte
+}
```
apatsap: We should use a `securemem.MemVault` here because then we don't need to care about re-implementint the cleanup and the wipeRegistry can go altogether

internal/kmip/get.go
apatsap: as @jithinkunjachan said we are potentially opening a leakage vector here.

A) by copying into a non secure mem
B) `if reg := wipeRegistryFromCtx(ctx); reg != nil` if req is nil we are not wiping key material


```go
secVault := memVaultFromCtx(ctx); reg != nil {
return error
}

mat, err := dek.Material.Clone("new name")
if err {
...
}

secVault.ImportData(mat)
```

apatsap (changes requested):
one thing that slipped through

internal/kmip/memkeymanager.go
NamanBalaji: This and vault is just mocks for now to show functionality and how the server will work once we have all the pieces

apatsap (approved):

PR #174, opened 2026-07-03
feat: add cleanup for data

jithinkunjachan:
Description
Add a runtime.AddCleanup safety net to securemem.Data that automatically destroys mmap'd memory regions when the GC collects an unreachable *Data instance — preventing memory leaks if a caller forgets to call Destroy().
Key design points:
- A dataCleanupRef struct captures only the slice header and name (not *Data), so it doesn't prevent GC collection of the parent object.
- The cleanup callback sets isReadOnly: true on a temporary Data, forcing Destroy() to call readwrite() (mprotect) first. This safely returns ENOMEM on already-unmapped memory instead of SIGSEGV from writing to invalid addresses.
- Explicit Destroy() calls cleanup.Stop() to deregister the callback, preventing double-destroy.
- Adds structured logging to the destroy lifecycle for observability.
- Subprocess-based integration tests verify both paths: cleanup fires when Destroy() is not called, and cleanup is suppressed when Destroy() is called explicitly.

### DIFF
--- a/internal/securemem/data.go
+++ b/internal/securemem/data.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"runtime"
 	"sync"
 )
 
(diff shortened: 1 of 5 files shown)


apatsap (commented):
Nice. And thx for the swift fix

apatsap (approved):

fabenan-f (approved):
looks good!

PR #184, opened 2026-07-17
feat: make key processor manager version aware

fabenan-f:
Introduce key versioning to the key processor by replacing `model.Key` with `model.KeyVersion` in all processor request types and adding version-aware resolution in the Manager, which now resolves the highest usable revision from the store before dispatching to processors


apatsap (approved):

NamanBalaji (approved):
Nice work

PR #185, opened 2026-07-21
Proposal: split up the Cryptor interface

fabenan-f:
The `cryptor.Cryptor` used for all three roles in the key processor is split into three purpose-specific interfaces:

- **`cryptor.Cryptor`** (`internal/cryptor/cryptor.go`): encrypts/decrypts using an externally provided secret and AAD
- **`cryptor.Sealer`** (`internal/cryptor/sealer.go`): seals/unseals with a self-managed key bound to a trust anchor; no external secret
- **`keyprocessor.SecretWrapper`** (`internal/keyprocessor/secret_wrapper.go`): wraps/unwraps secrets within a key hierarchy using a key-version path

A `RootSealerAdapter` bridges the root-HSM case where a `Sealer` must sit in the `Cryptor` slot.

**Benefits**

- Compile-time guarantees prevent plugging the wrong implementation into the wrong slot
- Each interface exposes only what its role requires; no optional/ignorable fields
- Method names reflect intent (`Seal` vs `Encrypt` vs `WrapSecret`), making the code self-documenting

**Drawbacks**

- `DecryptionSecretRequired` remains a runtime flag since root can use either a `Sealer`-backed or a full `Cryptor`
- A few breaking changes across existing implementations and callers due to the interface split

### DIFF
--- a/internal/cryptor/sealer.go
+++ b/internal/cryptor/sealer.go
@@ -1,0 +1,81 @@
+package cryptor
+
+import (
+	"context"
+	"errors"
+
+	"github.com/openkcm/krypton/internal/securemem"
+)
+
+type SealRequest struct {
+	Plaintext *securemem.Data
+	AAD       []byte
+}
+
+type SealResponse struct {
+	Ciphertext *securemem.Data
+}
+
+type UnsealRequest struct {
+	Ciphertext *securemem.Data
+	AAD        []byte
+}
+
+type UnsealResponse struct {
+	Plaintext *securemem.Data
+}
+
+// Sealer encrypts and decrypts with a self-managed key bound to a trust anchor.
+// No external secret is provided; the key material is intrinsic to the implementation.
+type Sealer interface {
+	Seal(ctx context.Context, req SealRequest) (*SealResponse, error)
+	Unseal(ctx context.Context, req UnsealRequest) (*UnsealResponse, error)
+	Info() Info
+}
+
+// ErrUnexpectedSecret is returned when a RootSealerAdapter receives a non-nil Secret,
+// indicating a wiring bug in the processor configuration.
+var ErrUnexpectedSecret = errors.New("root sealer adapter received unexpected secret")
+
+// RootSealerAdapter adapts a Sealer to the Cryptor interface for root keys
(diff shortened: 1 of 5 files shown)


internal/cryptor/sealer.go
```diff
@@ -0,0 +1,81 @@
...
+	"context"
+	"errors"
+
+	"github.com/openkcm/krypton/internal/securemem"
+)
+
+type SealRequest struct {
+	Plaintext *securemem.Data
+	AAD       []byte
+}
```
apatsap, line 13: The `TenantID`, `KeyID` and `KeyVersion` must be there.

internal/cryptor/sealer.go
```diff
@@ -0,0 +1,81 @@
...
+
+type SealResponse struct {
+	Ciphertext *securemem.Data
+}
+
+type UnsealRequest struct {
```
apatsap, line 19: The TenantID, KeyID and KeyVersion must be there.

internal/keyprocessor/processor.go
```diff
@@ -22,62 +22,62 @@ var (
...
 // It destroys all intermediate secrets that pass through its operations.
-type Processor struct {
+type processor struct {
 	generator       cryptor.SecretGenerator
-	wrapper         cryptor.Cryptor
-	transportSealer cryptor.Cryptor
-	parent          cryptor.Cryptor
+	cryptor         cryptor.Cryptor
+	transportSealer cryptor.Sealer
+	parent          SecretWrapper
```
apatsap, line 31: this becomes a sealer

internal/keyprocessor/secret_wrapper.go
```diff
@@ -0,0 +1,37 @@
...
+	WrappedSecret *securemem.Data
+	AAD           []byte
+}
+
+type UnwrapSecretResponse struct {
+	Secret *securemem.Data
+}
+
+// SecretWrapper wraps and unwraps secrets within a key hierarchy.
+type SecretWrapper interface {
```
apatsap, line 34: this can go

apatsap (commented):
`DecryptionSecretRequired` this should go now

PR #186, opened 2026-07-23
feat: add sealer interface

fabenan-f:
These changes introduce a clear separation between key-version-aware sealing and low-level encryption.

- Sealer interface for key-version-aware seal/unseal operations used by StaticSecret, Manager, and RootManager
- Cryptor.Secret is mandatory, so no more optional pointer or DecryptionSecretRequired flag
- Processor uses Sealer for transport sealing (optional) and parent delegation (mandatory)
- Manager validates key lifecycle, resolves key versions and secrets, then delegates encryption/decryption to processors.
- RootManager validates key lifecycle and delegates to a sealer for root keys that manage their own secret

### DIFF
--- a/internal/cryptor/sealer.go
+++ b/internal/cryptor/sealer.go
@@ -10,7 +10,7 @@ import (
 type SealRequest struct {
 	TenantID   string
 	KeyID      string
-	KeyVersion *string
+	KeyVersion string
 	Plaintext  *securemem.Data
 	AAD        []byte
 }
@@ -24,7 +24,7 @@ type SealResponse struct {
 type UnsealRequest struct {
 	TenantID   string
 	KeyID      string
-	KeyVersion *string
+	KeyVersion string
 	Ciphertext *securemem.Data
 	AAD        []byte
 }
--- a/internal/cryptor/sealer.go
+++ b/internal/cryptor/sealer.go
@@ -1,0 +1,41 @@
+package cryptor
+
+import (
+	"context"
+
+	"github.com/openkcm/krypton/internal/securemem"
+)
+
+// SealRequest identifies a key version and provides the plaintext and AAD to seal.
+type SealRequest struct {
+	TenantID   string
+	KeyID      string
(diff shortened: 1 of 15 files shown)


internal/keyprocessor/manager.go
```diff
@@ -15,88 +15,138 @@ var (
...
-		return nil, err
+		return cryptor.UnsealResponse{}, err
 	}
 
 	key, err := km.store.GetKeyByID(ctx, req.KeyID, req.TenantID)
 	if err != nil {
-		return nil, err
+		return cryptor.UnsealResponse{}, err
+	}
+	if key.LifeCycleState != model.KeyLifeCycleActive {
```
fabenan-f, line 89: For now only state `active` is allowed but this needs to be aligned with NIST in the future

internal/cryptor/sealer.go
```diff
@@ -0,0 +1,41 @@
...
+// SealResponse holds the sealed ciphertext.
+type SealResponse struct {
+	Ciphertext *securemem.Data
+}
+
+// UnsealRequest identifies a key version and provides the ciphertext and AAD to unseal.
+type UnsealRequest struct {
+	TenantID   string
+	KeyID      string
+	KeyVersion *string
```
apatsap: Key version isn't optional

apatsap (changes requested):
One thing

internal/cryptor/sealer.go
fabenan-f: It would be for root keys where we do not have versions, otherwise this needs to be clarified

apatsap: Fair enough. But even if it is optional i wouldn't make it a pointer since empty string conveys the same info and its much easier to work with in code

fabenan-f: I changed it

apatsap (approved):

NamanBalaji (approved):

PR #192, opened 2026-07-28
feat: add sealer config

fabenan-f:
This change separates sealer and cryptor into distinct configuration fields on KeyBinding, introduces a sealerprovider package, and validates that root keys only have a sealer, TEK keys have both, and all other keys have a cryptor.


internal/cryptor/sealer.go
fabenan-f, line 1: A proper `sealer` package might encapsulate this better

apatsap (approved):

jithinkunjachan (approved):

PR #193, opened 2026-07-29
feat: add ExportSecret and using it in KMIP get handler

NamanBalaji:


internal/keyprocessor/manager.go
```diff
@@ -25,38 +26,69 @@ var (
...
 var _ cryptor.Sealer = &Manager{}
 
-// Seal validates the key lifecycle, resolves the key version and its secret to encrypt the plaintext.
-func (km *Manager) Seal(ctx context.Context, req cryptor.SealRequest) (cryptor.SealResponse, error) {
-	if req.KeyVersion == "" {
-		return cryptor.SealResponse{}, ErrKeyVersionRequired
-	}
-	kv, err := km.resolveUsableKeyVersion(ctx, req.TenantID, req.KeyID, req.KeyVersion)
+// NewManager creates a Manager backed by the given stores and processors.
+func NewManager(s store.Key, kvs store.KeyVersion, processors map[model.KeyKind]*Processor) *Manager {
```
fabenan-f: ~~The idea was to introduce `map[model.KeyKind]processor` as its own type returned by `NewRepository(bindings map[model.KeyKind]spec.KeyBind, hierarchy spec.KeyHierarchy)`, which could then be passed directly to `NewManager()`, keeping `processor` unexposed.~~

~~But I guess for now we can leave it as is~~

I created a constructor that leaves the key processors themselves unexposed by using the key bindings and key hierarchy information

https://github.com/openkcm/krypton/pull/194

internal/keyprocessor/manager.go
```diff
@@ -25,38 +26,69 @@ var (
...
+	return &Manager{store: s, versionStore: kvs, processors: processors}
+}
+
+// ExportSecretRequest identifies the key whose plaintext material to export.
+type ExportSecretRequest struct {
+	TenantID string
+	KeyID    string
+	// KeyVersion is optional; when empty the latest usable version is resolved.
```
fabenan-f: imo we should work with explicit versions

internal/kmip/keymanager.go
```diff
@@ -18,15 +18,17 @@ var (
...
+	// ErrNotSupported signals that the KeyManager implementation does not
+	// support the requested operation.
+	ErrNotSupported = errors.New("kmip: operation not supported")
 )
 
-// KeyManager abstracts the retrieval of DEK material and metadata. This
-// ticket ships an in-memory fake implementation; a follow-up will provide a
-// real adapter backed by the SQL key store and the unwrap path.
+// KeyManager abstracts the retrieval of DEK material and metadata.
 type KeyManager interface {
```
fabenan-f: Do we expect an implementation other than the one krypton provides?

internal/kmip/keyprocessormanager.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/openkcm/krypton/internal/cryptor"
+	"github.com/openkcm/krypton/internal/keyprocessor"
+	"github.com/openkcm/krypton/internal/securemem"
+	"github.com/openkcm/krypton/pkg/model"
+	"github.com/openkcm/krypton/pkg/store"
+)
+
+// SecretExporter is the subset of *keyprocessor.Manager the KMIP adapter needs.
+type SecretExporter interface {
```
fabenan-f: Again do we expect an implementation other than the one krypton provides?

internal/kmip/memkeymanager.go
fabenan-f, line 1: Do we still need this mock?

internal/kmip/get.go
```diff
@@ -5,17 +5,19 @@ import (
...
-// backed by mlocked memory that the TerminateHook zeroes and unmaps when the
-// connection ends. If no vault is present the handler fails closed rather than
-// fall back to unlocked memory.
+// Key material handling: the caller-owned DEK material is imported into the
+// connection's securemem.MemVault, so the response payload is backed by
+// mlocked memory that the TerminateHook zeroes and unmaps when the connection
+// ends. If no vault is present the handler fails closed rather than fall back
+// to unlocked memory.
 func (h *handler) handleGet(ctx context.Context, req *payloads.GetRequestPayload) (*payloads.GetResponsePayload, error) {
 	tenantID, keyID, err := parseKeyIdentifier(req.UniqueIdentifier)
```
fabenan-f: Shouldn't we expect the entire key version path?

fabenan-f (commented):
A few questions/clarifications

internal/securemem/data.go
```diff
@@ -237,3 +237,14 @@ func (m *dataCleanupRef) Destroy() error {
...
+// Destroy zeroes and unmaps d, logging any failure. Safe to call with nil and
+// on already-destroyed data.
+func Destroy(d *Data) {
+	if d == nil {
+		return
+	}
+	if err := d.Destroy(); err != nil {
+		slog.Error("failed to destroy secure data", "name", d.Name(), "error", err)
+	}
+}
```
apatsap: I think having 2 destroy function/method is not necessary 

I'd add the `nil` check to the method and remove this function agan.

```go
func (m *Data) Destroy() error {
   if m == nil {
       return nil
  }
...
```

internal/kmip/keymanager.go
apatsap: I'd also just get rid of it for now since there are a lot of interfaces in the kmip pkg already

internal/kmip/keyprocessormanager.go
```diff
@@ -0,0 +1,94 @@
...
+	ExportSecret(ctx context.Context, req keyprocessor.ExportSecretRequest) (cryptor.Secret, error)
+}
+
+var _ SecretExporter = (*keyprocessor.Manager)(nil)
+
+// keyProcessorKeyManager adapts the key processor hierarchy to the KeyManager
+// port.
+type keyProcessorKeyManager struct {
```
apatsap: This reads not good. I think this should be just `KeyManager` (assuming we get rid of the interface)

internal/kmip/keyprocessormanager.go
apatsap: I think this is not about multiple implementations but rather to restrict the API to only what the KMIP server actually needs to operate.  

Question would be though, if this interface then should live here. I think it to move to `keyprocessor`

apatsap (changes requested):

internal/keyprocessor/manager.go
NamanBalaji: I have rebased on your branch

internal/kmip/keymanager.go
NamanBalaji: Removed

internal/kmip/keyprocessormanager.go
NamanBalaji: Also removed for now in future if we need to limit what KMIP user from the manager we can introduce it back

fabenan-f (approved):
lgtm

apatsap (approved):

PR #195, opened 2026-07-30
feat: change key version from string to int and add ORDER BY in Listkeyversion

jithinkunjachan:
feat: change key version from string to int and add ORDER BY in ListKeyVersions

This covers the two main changes:
* Refactoring Version/ParentKeyVersion from string to int across the entire stack (model, store, vault, cryptor, migrations, tests)
* Adding the ORDER BY clause so DESC applies to each column individually, and adding IsOrderByVersionDesc + LifeCycleState filter support to ListKeyVersions

### DIFF
--- a/pkg/store/keyversion.go
+++ b/pkg/store/keyversion.go
@@ -28,6 +28,7 @@ type KeyVersionOrder int
 
 const (
 	KeyVersionOrderUnspecified KeyVersionOrder = iota
+	KeyVersionOrderVersionDesc
 	KeyVersionOrderCreatedAtDesc
 	KeyVersionOrderRevisionDesc
 )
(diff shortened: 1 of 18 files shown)


pkg/store/keyversion.go
```diff
@@ -25,8 +25,10 @@ type CreateKeyVersionResult struct {
 type ListKeyVersionsQuery struct {
 	TenantID              string
 	KeyID                 string
-	Version               string
+	Version               int
+	LifeCycleState        model.KeyLifeCycleState
 	ProcessingState       model.KeyVersionProcessingState
+	IsOrderByVersionDesc  bool
```
fabenan-f: np: I find @NamanBalaji's approach here a bit cleaner: https://github.com/openkcm/krypton/pull/193/changes#diff-f7d1544232c4c40ce31e4342735fcd6223c0956308d2eb3a042bb782b0559185R40

fabenan-f (approved):
just one comment which may be considered, and thanks for the change!

apatsap (approved):

fabenan-f (approved):

fabenan-f (approved):

NamanBalaji (approved):
lgtm

PR #197, opened 2026-07-31
feat: add ValidateActivateKey to key validator

jithinkunjachan:
This adds activation validation that checks the key hierarchy order, parent key states, processing status, and lifecycle transition validity before allowing a key to be activated.

### DIFF
--- a/pkg/validator/key_validator.go
+++ b/pkg/validator/key_validator.go
@@ -60,7 +60,7 @@ type AnnounceInput struct {
 	TargetName string
 }
 
-func (v *keyValidator) ValidateAnnounceKey(ctx context.Context, input AnnounceInput) *ValidationError {
+func (v *keyValidator) ValidateKeyAnnounce(ctx context.Context, input AnnounceInput) *ValidationError {
 	ve := &ValidationError{
 		code: Invalid,
 	}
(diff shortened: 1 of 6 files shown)


pkg/validator/key_validator.go
```diff
@@ -105,6 +115,71 @@ func (v *keyValidator) ValidateAnnounceKey(ctx context.Context, input AnnounceIn
 	return nil
 }
 
+// ValidateActivateKey implements [KeyValidator].
+func (v *keyValidator) ValidateActivateKey(ctx context.Context, input ActivateInput) *ValidationError {
```
apatsap: NP can we call it `ValidateKeyActivation` and `ActivationInput`

I guess we would then also rename the `ValidateAnounceKey` to `ValidateKeyAnnouncment`

That would ultimately lead to 
ValidateKeyActivation
ValidateKeyDeactivation
ValidateKeyCompromise
ValidateKeyDestruction
ValidateKeySuspension
ValidateKeyRevocation

IMHO this read better

pkg/validator/key_validator.go
```diff
@@ -105,6 +115,71 @@ func (v *keyValidator) ValidateAnnounceKey(ctx context.Context, input AnnounceIn
...
+	ve := &ValidationError{
+		code: Invalid,
+	}
+
+	switch {
+	case input.TenantID == "":
```
apatsap: we need to actually validate the tenant ID. Since in line 133 we do an actual DB query without really vetting the TenantID

Idea we have  `TenantIDInput` etc. type which can be used accross different input requests which makes sure its not some bogus enumeration attack

pkg/validator/key_validator.go
```diff
@@ -105,6 +115,71 @@ func (v *keyValidator) ValidateAnnounceKey(ctx context.Context, input AnnounceIn
...
+		ve.err = ErrEmptyTenantID
+		return ve
+	case input.KeyID == "":
```
apatsap: Same as tenantID missing vetting

apatsap (changes requested):

apatsap (approved):

fabenan-f (approved):

PR #198, opened 2026-07-31
fix: escape sql LIKE/ILIKE wildcards

fabenan-f:
This fixes potential LIKE/ILIKE wildcard injection in ListKeys name filter by escaping %, _, and \ so user input is matched as a literal substring


apatsap (approved):
Thanks!

jithinkunjachan (approved):

PR #204, opened 2026-08-04
fix: decouple hierarchy from bindings in key processor manager

fabenan-f:
Decouple hierarchy from bindings in `keyprocessor.NewManager`: root-only and agent-only segments now both construct


jithinkunjachan (approved):

jithinkunjachan:
Thank you 🙏

apatsap (approved):

PR #205, opened 2026-08-05
feat: add activate-key CLI command and integration tests

jithinkunjachan:
Introduce the  subcommand to the CLI, wiring it to the existing ActivateKey RPC. 
Promote keyprocessor.NewManager initialization to shared infrastructure so it can serve both KMIP and the new CLI path. 
Refactor integration test helpers to return a testSetup struct, giving each test its own tenant for better isolation.

### DIFF
--- a/cli/cmd/activate.go
+++ b/cli/cmd/activate.go
@@ -1,0 +1,79 @@
+package cmd
+
+import (
+	"fmt"
+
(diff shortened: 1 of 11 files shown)


fabenan-f (approved):
lgtm

apatsap (approved):

PR #207, opened 2026-08-06
feat: implement kmip get attributes endpoint

NamanBalaji:


apatsap (approved):

apatsap (approved):

internal/keyprocessor/manager.go
```diff
@@ -69,6 +70,7 @@ func NewManager(ctx context.Context, cfg ManagerConfig) (*Manager, error) {
 		store:        cfg.KeyStore,
 		versionStore: cfg.KeyVersionStore,
 		processors:   make(map[model.KeyKind]processor, len(cfg.Bindings)),
+		algorithms:   make(map[model.KeyKind]cryptor.KeyAlgorithm, len(cfg.Bindings)),
```
fabenan-f, line 73: I was wondering if the algorithm should map per key (being a field in the model) rather than per key kind

fabenan-f (approved):
Just one thought I already had in mind which we can discuss or we just go with this for now

internal/keyprocessor/manager.go
NamanBalaji: Good point. Currently there's only AES256 supported so no way to have any other algorithm but Ideally we should have this field on the Key model. Can handle this in a new PR as it requires schema changes.

jithinkunjachan (approved):

PR #213, opened 2026-08-13
feat: added transactions for the database store

NamanBalaji:


pkg/store/transactor.go
```diff
@@ -0,0 +1,37 @@
...
+// on the Stores passed to fn joins that transaction. If fn returns an
+// error or panics the transaction is rolled back, otherwise it is
+// committed. A nested call opens an independent transaction. The Stores
+// passed to fn must not outlive fn or be shared across goroutines, as the
+// transaction they are bound to is not goroutine-safe.
+type Transactor interface {
+	Transaction(ctx context.Context, fn TransactionFunc) error
+}
+
+// TransactorFunc adapts a function to the Transactor interface, allowing
```
fabenan-f: is this used anywhere?

fabenan-f (approved):

pkg/store/sql/transactor.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+var _ store.Transactor = &Transactor{}
+
+func NewTransactor(db *sql.DB) *Transactor {
+	return &Transactor{db: db}
+}
+
+func (t *Transactor) Transaction(ctx context.Context, fn store.TransactionFunc) error {
+	tx, err := t.db.BeginTx(ctx, nil)
```
apatsap: I think this uses the defaul `read committed`

Potential problem is that we can end up with 2 different key versions.
 
I'd go with the highest isolation. 
```go
tx, err := t.db.BeginTx(ctx, &sql.TxOptions{
      Isolation: sql.LevelSerializable,
  })
```

pkg/store/sql/transactor_test.go
```diff
@@ -0,0 +1,152 @@
...
+				return stores.Keys.CreateKey(ctx, key)
+			})
+			if innerErr != nil {
+				return innerErr
+			}
+			return errBoom
+		})
+
+		// then: the inner transaction committed on its own, unaffected by
+		// the outer rollback
```
apatsap, line 125: I'd expect that also the inner gets rolled back, since its parent transaction failed. Whats the rational behind this implementation?

apatsap (changes requested):

pkg/store/sql/transactor.go
NamanBalaji: Good point, changed

pkg/store/sql/transactor_test.go
NamanBalaji: This test serves more like a caution, we should never nest transactions and I can't think of a case where we would need to nest them. Postgres itself does not support nested BEGIN statements so this test is basically a translation of that in our code.

pkg/store/transactor.go
NamanBalaji: No I forgot to clear this it's left from a previous impl

apatsap (approved):

jithinkunjachan (approved):

PR #214, opened 2026-08-14
feat: add mTLS authentication and refactor login into subcommands

jithinkunjachan:
- Add mTLS as an authentication provider with cert chain verification
- Refactor login into login no-auth and login mtls subcommands
- Centralize gRPC connection creation based on stored token type
- Extract reusable internal/tlsconf package from KMIP TLS config
- Enable optional server-side TLS via root config

Note: internal mTLS will be added in the subsequent PRs


internal/tlsconf/server.go
```diff
@@ -0,0 +1,26 @@
...
+func (cfg *Server) BuildTLSConfig() (*tls.Config, error) {
+	certs, pool, err := tlsConfig(cfg.Cert, cfg.Key, cfg.ClientCA)
+	if err != nil {
+		return nil, err
+	}
+	return &tls.Config{
+		Certificates: certs,
+		ClientAuth:   tls.RequireAndVerifyClientCert,
+		ClientCAs:    pool,
+		MinVersion:   tls.VersionTLS12,
```
apatsap: We should use 1.3

apatsap (changes requested):
## certificate cross validation
We need to check the agent name against the certificate in all methods in`pkg/api/v1/proto/agents/service.go` 
+ Register
+ Deregister
+ SendHeartbeat
```go
func (a *AgentService) Register(ctx context.Context, r *RegisterAgentRequest) (*RegisterAgentResponse, error) {
	agentName := r.GetAgentName()
        // Needs to be checked against cert
```


Similar problem in 
+ `pkg/api/v1/proto/admin/keys/key_service.go`
+ `pkg/api/v1/proto/admin/tenant_service.go`

## Other things

+ the agent seems to still use `insecure.NewCredentials()`

apatsap (approved):

fabenan-f (approved):

